PR #11, opened 2017-08-16
Adding myself to admins.

gunjanbaid:


datahub/config.yaml
```diff
@@ -23,3 +23,8 @@ singleuser:
     extraVolumeMounts:
       - name: home
         mountPath: /home/jovyan
+        
+admin:
+  users:
+    # infrastructure
+      - 'gunjanbaid'
```
yuvipanda: You don't need quotes around it

PR #27, opened 2017-08-20
Disable sqlite as provider of ipython history

yuvipanda:
We don't actually need it, and it puts an sqlite file on NFS which is meh.

### DIFF
--- a/user-image/Dockerfile
+++ b/user-image/Dockerfile
@@ -23,6 +23,8 @@ RUN echo "en_US.UTF-8 UTF-8" > /etc/locale.gen && \
 ENV LC_ALL en_US.UTF-8
 ENV LANG en_US.UTF-8
 ENV LANGUAGE en_US.UTF-8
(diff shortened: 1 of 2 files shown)


user-image/Dockerfile
```diff
@@ -66,4 +66,10 @@ RUN pip install git+https://github.com/data-8/nbresuse.git@2f9144f && \
 RUN pip install git+https://github.com/data-8/gitautosync@b32a087 && \
 	jupyter serverextension enable  --sys-prefix --py nbgitautosync
 
+# disable history.sqlite
+ENV IPYTHONDIR ${APP_DIR}/venv/etc/ipython
+RUN mkdir -p ${IPYTHONDIR}/profile_default && \
```
yuvipanda: I think you should make this an actual file and do an ADD, rather than echo.

yuvipanda: I made https://github.com/berkeley-dsep-infra/datahub/pull/28

Issue #33, opened 2017-08-22
Pre-pull data8 materials


yuvipanda:
To reduce io hit on day 1 of class, let's do the following:

1. Gather list of all users enrolled for data8
2. Find their usernames on disk by using the following code snippet:
``` 
        import strings
        import escapism
        username = 'username' # Only part before @
        safe_chars = set(string.ascii_lowercase + string.digits)
        safe_username = escapism.escape(username, safe=safe_chars, escape_char='-').lower()
```
3. Copy a clone of the repo to /data/homes/{safe_username}

yuvipanda:
we can test this by getting a user who has any special chars in their username to login and verify a clone is already there.

yuvipanda:
Also we must make sure origin for all these is the appropriate upstream repo - https://github.com/data-8/materials-fa17. We can just make one clone and just cp -r it everywhere.

yuvipanda:
We did this!

PR #35, opened 2017-08-22
Add pre-puller.

ryanlovett:
First pass at #33.

### DIFF
--- a/pre-pull.py
+++ b/pre-pull.py
@@ -1,0 +1,68 @@
+#!/usr/bin/python
+
+import argparse
+import os
+import sys
+import string
+import subprocess
+
+import escapism
+
+safe_chars = set(string.ascii_lowercase + string.digits)
+repo = 'https://github.com/data-8/materials-fa17.git'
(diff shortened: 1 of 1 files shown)


pre-pull.py
```diff
@@ -0,0 +1,68 @@
...
+repo_dirname = os.path.basename(args.repo).split('.')[0]
+
+if not os.path.exists(local_repo):
+    os.mkdir(local_repo)
+
+git_clone()
+
+f = open(args.filename)
+line = f.readline()
+while line != '':
```
yuvipanda: awesome. did you already run this? :)

ryanlovett: Yep, last night. I tested it with myself but as you noted it'd be good to test with a username that requires massaging.

yuvipanda:
Let's not call this 'pre-puller' since there's already an image pre-puller. Can you make a dir called scripts and call this 'git-pre-cloner' or something?

ryanlovett:
Replaced by #46.

PR #36, opened 2017-08-22
Added networkx for social networks connector.

gunjanbaid:


henchc:
looks like you forgot an equals sign, i also accidentally didn't read through the pull request and added it along with bokeh in the PR above.

PR #40, opened 2017-08-23
Updating scipy, adding nano, vim, and gfortran for phys151.

gunjanbaid:
Updating from 0.19.0 to 0.19.1. This is to deal with memory leaks in `quad()`, described [here](
https://github.com/scipy/scipy/issues/7214). 
Also added vim, nano, and gfortran. 
@rainwoodman

### DIFF
--- a/user-image/Dockerfile
+++ b/user-image/Dockerfile
@@ -14,7 +14,11 @@ RUN apt-get update && \
             wget \
             npm \
             nodejs-legacy \
-            locales && \
+            locales \
+            nano \
+            vim \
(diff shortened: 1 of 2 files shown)


rainwoodman:
Thanks. When will the changes be pushed to the machine?

yuvipanda:
@rainwoodman it's in staging.datahub.berkeley.edu now, and I tested it and it seems ok. It'll be in datahub.berkeley.edu as soon as https://github.com/berkeley-dsep-infra/datahub/pull/41 is merged.

PR #41, opened 2017-08-24
Merge #40 to prod

yuvipanda:


rainwoodman:
Thank you all for the quick response. I can confirm the scipy leak is fixed.

PR #65, opened 2017-09-09
Disable ECN prior to docker builds.

ryanlovett:
https://github.com/travis-ci/travis-ci/issues/2389

We can remove this if pip is ever interrupted by network issues during builds.


deploy.py
```diff
@@ -24,6 +24,12 @@ def build_user_image(image_name, commit_range=None, push=False):
             print("user-image not touched, not building")
             return
 
+    # Attempt to improve relability of pip installs:
+    # https://github.com/travis-ci/travis-ci/issues/2389'''
+    subprocess.check_call([
```
yuvipanda: This is travis specific, and should hence be in the travis specific `travis/travis-script.bash` file. `deploy.py` should be able to run anywhere.

ryanlovett:
Obviated by #66.

PR #95, opened 2017-09-22
ENH add pycortex filestore

eickenberg:
Hi everybody,

this PR was supposed to add all the data necessary for the cogneuro connector this semester, but I have run into some git/github issues and have decided to chunk this and add only the most urgent files.

**Important Note** Even only this PR will add several hundred MB to the repo, which is currently only at around 60MB in size. If I manage to add the rest of the data, this will inflate to more than 2.5GB. This may be **very annoying** for everybody else who works with this repo to download. So I was wondering whether there is any other way to make data sets available to the students.

The files in the current PR are necessary for next Monday's lecture.

Thanks!
Michael


gunjanbaid:
Hi @eickenberg! Thanks for pointing this out. This was my bad, I thought this repo would be the place to add data files, but we definitely don't want 2GB in here. So sorry about the confusion. Instead, I can copy the files to the NFS server and mount that onto /data/cogneuro on datahub.

**To do so, I'll need a link from which to wget the files. Could you upload the files to box/dropbox/etc. and give me the URL?** There's not a good way to `wget` the files from drive.

Once I have a URL, I'll add the files to the NFS server used for datahub and update the configs. You'll be able to access the files on the hub at /data/cogneuro. We have a few threads going for this topic, so let's follow up here! Adding @samyag1 to this thread also. Thank you!

gunjanbaid:
These changed requested in this PR were added in https://github.com/berkeley-dsep-infra/datahub/pull/97 and https://github.com/berkeley-dsep-infra/datahub/pull/98, so closing this.

PR #100, opened 2017-09-25
Updating pycortex, adding needed packages for cogneuro connector.

gunjanbaid:


user-image/connectors/neuro.bash
```diff
@@ -11,9 +11,9 @@ pip install --no-cache-dir cython==0.26
 # ${CONDA_DIR}/bin/pip install git+https://github.com/gallantlab/pycortex@data8
 
 # Installs the cortex library
-git clone https://github.com/gallantlab/pycortex
+git clone https://github.com/eickenberg/pycortex.git
 cd pycortex
 # commit on data8 branch
-git checkout d570713
+git checkout data8
```
yuvipanda: I'd prefer this be a hash instead of a branch, so we can force a rebuild when it changes.

gunjanbaid: Is commit hash ok? Was not sure what the previous hash (d570713) was.

yuvipanda: Yup, commit hash is ok!

yuvipanda:
lgtm! merge when you've time to take it through the whole process :)

PR #106, opened 2017-09-26
add basemap and dependencies

henchc:


user-image/Dockerfile
```diff
@@ -22,6 +22,10 @@ RUN apt-get update && \
             texlive-xetex \
             texlive-fonts-recommended \
             texlive-generic-recommended \
+	    # for LS88-5 and modules basemap
+	    libgeos-3.3.3 \
```
ryanlovett: Is libgeos-3.3.3 available in Ubuntu 17.04? (top line of Dockerfile)

I think this is the geos from Ubuntu 17.04:
https://launchpad.net/ubuntu/+source/geos/3.5.1-3

user-image/Dockerfile
```diff
@@ -22,6 +22,10 @@ RUN apt-get update && \
...
+	    libgeos-c1 \
```
ryanlovett: In 17.04 is it called libgeos-c1v5?

PR #109, opened 2017-09-29
Increase RAM for chench students.

ryanlovett:


yuvipanda:
This should include the changes in prod.yaml too

PR #114, opened 2017-10-02
Updating pycortex installation.

gunjanbaid:


gunjanbaid:
@eickenberg @samyag1 The pycortex updates should be available on datahub, after you do a server restart.

PR #115, opened 2017-10-03
adding folium

TomPiona:
Adding `folium` to requirements.txt for mapping capabilities. `datascience` uses an older version of folium, but in order to run straight `folium`, it needs the update to version 0.5.0.


henchc:
just to be clear, this is for modules!

henchc:
should be `==` @TomPiona can you add that?

PR #116, opened 2017-10-03
Revert "adding folium", breaks datascience.maps

henchc:
Reverts berkeley-dsep-infra/datahub#115

methods used in the data8 textbook were renamed in the newer `folium` version.

@vinitra could you look into maybe updating the `folium` version in the `datascience` dependencies?


vinitra:
Looking into this.

PR #121, opened 2017-10-06
add curl and pygame for modules

henchc:


user-image/Dockerfile
```diff
@@ -25,6 +25,8 @@ RUN apt-get update && \
 	    # for LS88-5 and modules basemap
 	    libspatialindex-dev \
 	    libgeos-dev \
+	    # for module
```
yuvipanda: would be nice to specify for which module :)

yuvipanda:
ty

PR #123, opened 2017-10-07
Adding neurodebian, fsl, nipype for cogneuro. 

gunjanbaid:
* Adding the `neurodebian` repo for apt based on instructions [here](http://neuro.debian.net/)
* Adding `python-nipype` based on instructions [here](http://neuro.debian.net/install_pkg.html?p=python-nipype)
* Adding `fsl-complete` based on instructions [here](http://neuro.debian.net/install_pkg.html?p=fsl-complete)

Moved around commands to make sure that `wget` is installed before adding `neurodebian`.

@samyag1 Could you confirm that these look good?
@yuvipanda Any changes needed?

CC @eickenberg


user-image/Dockerfile
```diff
@@ -28,9 +28,18 @@ RUN apt-get update && \
...
+            gfortran
+
+# for cogneuro connector, adding NeuroDebian repo
+RUN wget -O- http://neuro.debian.net/lists/zesty.us-ca.libre | tee /etc/apt/sources.list.d/neurodebian.sources.list
+RUN apt-key adv --recv-keys --keyserver hkp://pool.sks-keyservers.net:80 0xA5D32F012649A5A9
 
+RUN apt-get install --yes
+	    # for cogneuro connector
+	    fsl-complete \
+	    python-nipype && \
```
yuvipanda: This is probably not going to have any effect - installing python packages from apt is not going to be visible inside our virtual environment. They will need to be installed via pip.

user-image/Dockerfile
```diff
@@ -28,9 +28,18 @@ RUN apt-get update && \
...
             # for phys 151
-            gfortran && \
-    apt-get purge && apt-get clean
+            gfortran
+
+# for cogneuro connector, adding NeuroDebian repo
+RUN wget -O- http://neuro.debian.net/lists/zesty.us-ca.libre | tee /etc/apt/sources.list.d/neurodebian.sources.list
+RUN apt-key adv --recv-keys --keyserver hkp://pool.sks-keyservers.net:80 0xA5D32F012649A5A9
 
+RUN apt-get install --yes
```
yuvipanda: This line is missing a '\' at the end causing the build to fail too

user-image/Dockerfile
```diff
@@ -28,9 +28,18 @@ RUN apt-get update && \
 	    # for soc-130ac mapping module
 	    curl \
             # for phys 151
-            gfortran && \
-    apt-get purge && apt-get clean
+            gfortran
+
+# for cogneuro connector, adding NeuroDebian repo
+RUN wget -O- http://neuro.debian.net/lists/zesty.us-ca.libre | tee /etc/apt/sources.list.d/neurodebian.sources.list
```
yuvipanda: This is probably more clear as:

```
wget http://neuro.debian.net/lists/zesty.us-ca.libre  -O /etc/apt/sources.list.d/neurodebian.sources.list
```

they both have the same effect

user-image/Dockerfile
yuvipanda: You need to do an 'apt-get update' before this if you want the packages from the newly added repo to be available to apt.

user-image/Dockerfile
gunjanbaid: Ah ok. `fsl-complete` is not a python package, so I'll leave that there. I can use pip to install `nipype`. Initially using apt because @samyag1 had some concerns about version issues. @samyag1, does `pip install nipype==0.13.1` work here?

samyag1:
Works for me Gunjan, thanks for the quick reply!

samyag1:
Also, I didn't confirm, but those are the correct references for
installation, so we should be good. Thanks!

Sam

PR #136, opened 2017-10-21
Adding packages for smart cities, ce88.

gunjanbaid:


henchc:
@gunjanbaid i think Cython needs to be fully installed before `hmms`, so you could move `hmms` to the cogneuro bash file that gets run in the docker file after the requirements.txt .

So: Cython should be installed in req, hmms should be installed in neuro bash, i think.

gunjanbaid:
@henchc I did try that ordering but it didn't work unfortunately. But I'll try another variation of it, with cython installed earlier in the dockerfile.

henchc:
@gunjanbaid i think when you tried earlier it was still the cython 0.26?  now it looks like numpy is the problem :(

PR #137, opened 2017-10-22
Adding packages for ce88, updating cython version.

gunjanbaid:
@samyag1 @eickenberg Is version 0.25.2 of cython ok for pycortex? Previously, we were using version 0.26 of cython, but another course will need to use 0.25.2. Thanks!

### DIFF
--- a/user-image/connectors/neuro.bash
+++ b/user-image/connectors/neuro.bash
@@ -4,10 +4,6 @@ set -euo pipefail
 # nbserver; installed via requirements.txt
 jupyter serverextension enable  --sys-prefix --py nbserverproxy
 
-# pycortex seems to need cython, but isn't explicitly defined as a dependency
-# pip install --no-cache-dir cython==0.26
-# cython 0.25.2 installed in Dockerfile		--Gunjan
-
 # pycortex can't be installed from the repository at the moment
 # ${CONDA_DIR}/bin/pip install git+https://github.com/gallantlab/pycortex@data8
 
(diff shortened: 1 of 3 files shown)


gunjanbaid:
@yuvipanda @ryanlovett Do these changes look ok? Will wait to merge until you have approved and cogneuro instructors have confirmed that the cython version change is ok.

user-image/connectors/neuro.bash
```diff
@@ -5,7 +5,8 @@ set -euo pipefail
 jupyter serverextension enable  --sys-prefix --py nbserverproxy
 
 # pycortex seems to need cython, but isn't explicitly defined as a dependency
-pip install --no-cache-dir cython==0.26
+# pip install --no-cache-dir cython==0.26
+# cython 0.25.2 installed in Dockerfile		--Gunjan
```
yuvipanda: Don't leave in commented out code, remove it. And add a comment in where cython is being installed that it's used for multiple classes.

yuvipanda:
One comment, but otherwise if it builds + testing on staging looks good I'm ok. Thank you for pinging me!

ryanlovett:
Lgtm. It is totally beyond the scope of this issue, but you might consider sending a patch to (or filing an issue with) pycortex to add cython as an explicit dependency.

gunjanbaid:
Removed the commented code, as requested. Also got confirmation from cogneuro instructors that version 0.25.2 of cython is ok to use. If build succeeds, will merge this into staging. Thank you!

PR #169, opened 2017-12-24
Update travis-script.bash (staging)

jameswinegar:
remove duplicate line


yuvipanda:
Thanks for the patch, @jameswinegar!!!

PR #181, opened 2018-01-04
Add customization of docker images.

ryanlovett:


ryanlovett:
I need to abort this to pull in an additional change.

PR #205, opened 2018-01-10
 Updated admins for sp18

AdeelCheema:


ryanlovett:
@AdeelCheema do you want to merge this? Its okay to merge your own PRs if you don't need any feedback on them. I think you need to accept a repo invite though -- I can resend if needed.

AdeelCheema:
@ryanlovett Sounds good, thank you! I just accepted the invite and will merge this in.

AdeelCheema:
Actually, looks like I don't have write access, is there anything else that needs to be done besides accepting the invite?

ryanlovett:
@AdeelCheema I've just set you to be an admin on this repo so you should be able to merge.

PR #214, opened 2018-01-18
Added pysal for geog88

samyag1:
Added the following line per David O'Sullivan's request:
pysal==1.14.3


ryanlovett:
Fixed by #216.

PR #215, opened 2018-01-18
Use Ubuntu old-releases archive.

ryanlovett:


ryanlovett:
The sources.list change did not go through properly. Closing.

PR #222, opened 2018-01-18
Use extraEnv to hide username map in our secrets.

ryanlovett:


datahub/config.yaml
```diff
@@ -33,6 +34,14 @@ hub:
     # HACK: manually edited deployment to add volume to /srv/log
     c.JupyterHub.extra_log_file = '/srv/log/jupyterhub.log'
 
+    if 'HUB_USERNAME_MAP' in os.environ:
```
yuvipanda: Add a note saying this is a temporary fix that should go away soon?

ryanlovett: Done!

ryanlovett:
Closed in favor of #223.

PR #263, opened 2018-02-22
bump RAM for LS88-1 and PSYCH 88

henchc:


henchc:
@ryanlovett @AdeelCheema did this work? I just checked on staging:
```
$ pip freeze | grep datascience
datascience==0.10.1
```

ryanlovett:
@henchc Yes, it works here. Did you restart your server?

Users who are on the geog image won't see the update since the geog image didn't change. I need to modify deploy.py to rebuild the geog image whenever the base image changes.

henchc:
thanks @ryanlovett , i added myself to the geo image so that must be why! i guess we should wait to merge until tonight now.

PR #274, opened 2018-03-20
Adding scratchpad nbextension

AdeelCheema:


user-image/d8extension.bash
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+cd tmp
```
ryanlovett: This should be /tmp.

I'd also add `set -e` as the second line in the script so that the script fails if any line in the script errors.

AdeelCheema:
I do see it on staging, interesting that it worked with the tmp issue. I just pushed a change that should fix that. It should be accessible with ctrl + b, or the little arrow at the bottom right on staging.

ryanlovett:
Aah, it is because my account is associated with the geog88 image! I've verified that it works with another account of mine. Sorry for the noise!

AdeelCheema:
Alright, everything looks good to me, let me know if anything should be changed, otherwise, let's aim to merge in over the weekend/break.

ryanlovett:
Sounds good. I added a couple of users to the geo image too.

PR #277, opened 2018-04-23
Add apt-get imagemagick to dockerfile

eickenberg:
In order for pycortex to be able to render SVG-based regions of interest on flatmaps, it turns out that we need the imagemagick `convert` binary installed. I added a line to install all of imagemagick. 

Is there any chance for this to be staged + added to production by tomorrow? (That is the date of our last cogneuro connector class and it would be amazing if we could have this functionality for it).

Thanks!
Michael

### DIFF
--- a/user-image/Dockerfile
+++ b/user-image/Dockerfile
@@ -35,7 +35,7 @@ RUN apt-get update && \
             # for phys 151
             gfortran \
 	    # for cogneuro
-	    dirmngr
+	    dirmngr \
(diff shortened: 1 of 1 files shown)


ryanlovett:
@eickenberg @geoffbacon Am merging this to staging now. Should be easy to test, after which I'll PR/merge to prod.

ryanlovett:
We need to downgrade tornado to < 5 before we can merge to prod.

PR #279, opened 2018-04-24
Add imagemagick. Fix tornado at < 5.

ryanlovett:


eickenberg:
Thanks for the merge!!

That makes pycortex be able to render svg-based images in matplotlib!

I should add for precision that what we need for pycortex is the `convert` function from imagemagick. This specific function is also shipped with e.g. anaconda and to be found in `anaconda/bin` when installed. If the tornado downgrade is too much of a downgrade for other parts of datahub, then maybe there are other ways of only having

eickenberg:
cc @samyag1 for reference

ryanlovett:
@eickenberg The tornado change is actually necessary to make the jupyter notebook itself work properly on datahub. We had been using tornado 4.5.x on datahub but had not been specifying it as a pip requirement since it gets pulled in as a core dependency. If one doesn't specify the precise version, the latest version will get installed and tornado 5 has some breaking changes. I noticed the breakage when testing on staging.datahub.

PR #333, opened 2018-08-23
Add libraries for CP255 attempt #2

geoffbacon:
This is the second attempt to add libraries for CP255. I've removed what seemed to be the two libraries that caused problems last time.


ryanlovett:
Hey @geoffbacon , looks like this failed.

https://circleci.com/gh/berkeley-dsep-infra/datahub/95

The compiler can't find the gdal headers because the compilation is missing -I/usr/include/gdal. You might try
```
export CPLUS_INCLUDE_PATH=/usr/include/gdal
export C_INCLUDE_PATH=/usr/include/gdal
```
preceding the failing library in the cp255 script. Alternatively, you can try
`pip install gdal==2.2.4 --global-option=build_ext --global-option="-I/usr/include/gdal/"`

PR #343, opened 2018-09-10
Add prob140 deployment.

ryanlovett:


deployments/prob140/config/common.yaml
```diff
@@ -0,0 +1,73 @@
+jupyterhub:
+  scheduling:
+    userScheduler:
+      enabled: true
+  proxy:
+    https:
+      letsencrypt:
+        contactEmail: rylo@berkeley.edu
+  hub:
+    extraConfig: |
```
yuvipanda: This can move to hub/values.yaml I think, since it's common to datahub & prob140?

deployments/prob140/image/requirements.txt
```diff
@@ -0,0 +1,25 @@
...
+nbgitpuller==0.6.1
+nbzip==0.0.4
+jupyterlab==0.33.11
+ipywidgets==7.2.1
+jupyterhub==0.8.1
+appmode==0.4.0
+nbresuse==0.3.0
+#
+# r
+nbserverproxy==0.8.3
```
yuvipanda: we don't have rstudio here, does prob140 use R? if not we should remove it.

deployments/prob140/config/common.yaml
ryanlovett: Done. Since this touches the datahub deployment can you double-check my hub/values.yaml?

deployments/prob140/image/requirements.txt
ryanlovett: Fixed! Thanks, I'd removed the R bits from the user image but missed these.

PR #362, opened 2018-10-03
Update requirements.txt to address needs of EEP153/MinCostDiet.

ligon:
Would appreciate review of this---I'm not comfortable upgrading requirements.txt for everyone.


rameshvs:
I don't see any changes that look like they would break anything in the [numpy 1.15 release notes](https://github.com/numpy/numpy/blob/master/doc/release/1.15.0-notes.rst) or the [scipy docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html) for the functions we use. Still, with this big of a change, I'd prefer to test things out on a staging environment first to make sure all the lectures and assignments look okay. Would that be possible?

Thoughts, @papajohn or @davidwagner?

ryanlovett:
@rameshvs The easiest way to test in the short term would be to login to staging.datahub.berkeley.edu and run `pip install numpy==1.15.0 pandas==0.23.3 scipy==1.1.0`. This would be local to your environment there. After testing if you felt confident that we could go forward, we'd merge this to staging branch for broader testing.

rameshvs:
Will the new versions be installed through `pip` or `conda`? I noticed some warnings about numpy binary sizes after installing with pip and running through some of the materials. Looking at [this thread](https://github.com/ContinuumIO/anaconda-issues/issues/6678), it seems like the warnings can safely be ignored, but if possible, it would improve the student experience to not have them.

ryanlovett:
@login Could you update requests through a separate PR? I think that could go in right now because we know it won't break anything. Thanks for spotting it by the way!

ryanlovett:
Closing in favor of #489.

PR #367, opened 2018-11-01
New packages

ligon:
@ryanlovett Here's a new PR that includes only two additions to requirements.txt.

Versions are specified to avoid a security issue with version 2.18 of requests.


ligon:
I don't think that went  the way I meant for it too!  I'd intended only a single commit in  the PR.  

Let me try to figure out what I did wrong, and please don't merge this PR.

PR #441, opened 2019-01-17
Merge astr 128 libraries into prod

yuvipanda:


yuvipanda:
@kareemelbadry if you go to datahub.berkeley.edu now (and if necessary, restart your server with the 'control panel' button), you should have access to the libraries you added. Can you test and lmk if it works?

kareemelbadry:
Everything looks good. Thanks!

PR #458, opened 2019-01-22
Add very basic data8x deployment

yuvipanda:
data8x lives in a different cluster & project, but
no reason to have its own repository! 

- [x] PR Image building
- [x] Auth creds for GCR
- [x] Auth creds for GKE
- [ ] initContainer for volume ownership
- [ ] Create Cluster
- [ ] Create Filestore
- [ ] NFS mount options


yuvipanda:
Let's see if the pushing works!

PR #478, opened 2019-01-23
Introduce hubploy.yaml

yuvipanda:
- Specify image names for each deployment
- Make hubploy do auth for registry


yuvipanda:
meh, circleci restarted the run as it was about to finish. Boo!

PR #506, opened 2019-01-28
Force downgrade nbconvert

dibyaghosh:


dibyaghosh:
Hey!

nbconvert 4.x is the last versions of nbconvert in which converting notebooks to PDFs displays errors when the Latex fails to compile instead of silently aborting the compilation, and having a half-rendered PDF that includes no cell beyond the one with failed LaTex. 

As 140's grading pipeline involves students exporting their notebooks to PDFs, it's rather critical that students be made aware when their submission isn't full. I'm not aware of any way to enable this behavior in nbconvert >= 5.

yuvipanda:
@dibyaghosh there seems to be no way to use nbconvert < 5 with the current version of notebook we have either. We might have to try going back notebook versions slowly to see if any version works.

In the meantime, @ryanlovett is trying to fix this in nbconvert https://github.com/jupyter/nbconvert/pull/947

ryanlovett:
Fixed by 07f5088ff66719e4aa50334979e3a623daed7a05 and a7ce882d2b837b8cc5afc7487db511e1cdbbc2d4.

PR #511, opened 2019-01-29
Give astro 128 GSI additional RAM for testing.

ryanlovett:


kareemelbadry:
Hi, my calnet id is kelbadry

Thanks.

PR #574, opened 2019-02-22
howto switch over a hub to a new cluster

felder:


felder:
@yuvipanda please review the cleaned up documentation

yuvipanda:
@felder this looks good to me! Unfortunately, we're using restructuredText (www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html) format instead of markdown here for documentation. Can you convert this doc to rst? There's some automatic conversion help here: https://pandoc.org/try/

PR #631, opened 2019-04-12
Add finances doc.

ryanlovett:


docs/finances.rst
```diff
@@ -0,0 +1,29 @@
...
+Finances
+========
+
+Resource consumption and spending can be tracked and managed at our cloud
+providers and we log this usage in a Google doc.
+
+Google Cloud
+============
+
+Deployment assets are organized into projects and those projects have linked billing accounts. A project can be linked to only one billing account, however a billing account can be linked to more than one project. The accounts may be backed by grants or credit cards and can be switched out when needed. The current billing account runs through 1/2020.
```
yuvipanda: can we split the 'current billing account' into a separate subheading, and list it as:

active-usage-start-date | billing expiry date | billing account name

and we can update that as time goes?

ryanlovett: I just added a heading and table.

I don't know when we started using the previous two billing accounts.

PR #633, opened 2019-04-15
Update Gofer Submit

Yanay1:
Update Gofer Submit for multiple course support on the same data hub.


yuvipanda:
@Yanay1 thanks for the PR! I'll merge and see it through now, should be live on data8xhub shortly.

PR #654, opened 2019-04-28
spark gcs test

jameswinegar:


jameswinegar:
PR is focused on getting spark installed from binaries instead of from pip. This will let us pods as a cluster and let us integrate with GCS.

PR #655, opened 2019-04-28
spark with working gcs and local hadoop

jameswinegar:


deployments/w261/image/postBuild
```diff
@@ -9,22 +9,56 @@ cp ipython_config.py ${CONDA_DIR}/etc/ipython/ipython_config.py
 # Allow users to stop / start servers from JupyterLab
 jupyter labextension install @jupyterlab/hub-extension
 
-# Install Spark
-SPARK_HOME=/home/jovyan/spark
+# Download Spark
+SPARK_HOME=${HOME}/spark
```
yuvipanda: Since we can't use $HOME, let's try using $CONDA_DIR for now - I think it's writeable by the user.

PR #666, opened 2019-05-01
Add cookiecutter + instructions on adding new hub

yuvipanda:


yuvipanda:
The keys that were accidentally pushed here have been revoked.

PR #679, opened 2019-05-04
Update gofer submit and grader versions

Yanay1:


yuvipanda:
Thanks for opening the PR, @Yanay1!

Issue #725, opened 2019-05-22
Cleanup datahub image


yuvipanda:
The datahub image is too big and too unweildy. It also has packages and images
that are no longer required for classes that are no longer taught. Let's clean it up.

Classes currently with lots of special handling:

- [ ] CogNeuro. Not offered since Spring 2018 (https://classes.berkeley.edu/search/class/cogneuro?retain-filters). Uses NeuroDebian
- [ ] cp255. Lots of Geo stuff, many disabled. Last offered Spring 2017. https://classes.berkeley.edu/search/class/cp255?retain-filters=1
- [ ] ds100. Has its own hub.

ryanlovett:
I've checked with the prof for stat 159/259 about their dependencies.

yuvipanda:
@ryanlovett for R, I'm thinking of doing the following:

1. Remove all apt packages
2. Use repo2docker to get R / IRkernel
3. Install individual packages folks want with install.R

Do you know where the current list of R packages came from?

ryanlovett:
@yuvipanda Using install.R will be much slower because it has to compile every package. This would also require installing additional build dependencies (e.g. -dev packages) into the image though that's less of an issue. Personally I'd rather stick to apt packages if possible. One reason for not using apt is if the package version is too old.

All R packages were requested by @ericvd-ucb on behalf of instructors. Some dependencies of those packages were explicitly specified as apt packages where possible, to reduce build times.

ryanlovett:
@yuvipanda From experience I'd guess that using install.R for most R packages would add about ~ 20 min per build. It'd be worth checking though if/when you create the PR.

yuvipanda:
@ryanlovett this is tempting me to make a new R specific hub, that we can fold into a different image eventually.

ryanlovett:
Short term it'd require less development than adding a second image to the hub and enabling interact links to invoke the image. The only downside is that people might get used to the temporary R-based hub name.

I think the current classes that use python don't also use R, and vice versa. By the time we're handling those courses we'll be able to accommodate them with custom images.

yuvipanda:
@ryanlovett yeah, I like it. I'm a little torn on apt vs install.R, but we can probably move that for later.

ericvd-ucb:
[...]

PR #728, opened 2019-05-22
Remove ds100 packages

yuvipanda:
There's a data100 hub now

Ref #725


yuvipanda:
@ryanlovett can you confirm this is ok?

ryanlovett:
Totally fine to do this. The only reason I kept it in because I at some point I wanted to test whether wkhtmltopdf could generate PDFs more reliably than nbconvert. That's what Sam was using it for.

yuvipanda:
Thanks, @ryanlovett

PR #768, opened 2019-06-20
Bump JupyterHub to 1.0

yuvipanda:
- Brings with it a lot of new features & bugfixes - https://blog.jupyter.org/announcing-jupyterhub-1-0-8fff78acad7f
- Moves us forward, so we can keep up with changes better

This is using today's master version of the chart.


yuvipanda:
I tested this out on datahub staging and it worked out ok. I am gonna try do this either early tomorrow morning or late tomorrow night...

PR #775, opened 2019-06-21
Run R scripts with Rscript.

ryanlovett:


deployments/r/image/Dockerfile
```diff
@@ -29,6 +29,19 @@ RUN mkdir -p ${VENV_DIR} && chown -R ${NB_USER} ${VENV_DIR}
...
 
 USER ${NB_USER}
+
+# Install some R libraries that aren't in the base image
+COPY extras.d  /tmp/extras.d
+# CircleCI stops printing output at 40k chars.
+# We send stdout to a log file, and tail it a bit
+# FIXME: Find something less sucky
+# another hubploy #1
+RUN run-parts --list --regex '^.*.r$' /tmp/extras.d | \
```
yuvipanda: TIL about run-parts! Is there a reason we're usint it instead of ls? Does it require scripts be marked executable?

ryanlovett: It does. run-parts is basically for when you've got a bunch of executables which is kind of what this is for. If we put hashbangs in the scripts and make them +x then we can get rid of --list and xargs. But if you don't want that then we can just use ls.

Issue #882, opened 2019-07-23
Admin access for PH250WF+PH250WG instructors


anna-nguyen:
Can we add instructors for PH 250WF/G as admins on datahub?

**Instructor Emails:**
jadebc@berkeley.edu
jcolford@berkeley.edu
anna-nguyen@berkeley.edu
nolanpokpongkiat@berkeley.edu
wmjilek@berkeley.edu
stephaniemholm@gmail.com
lucascarlton@berkeley.edu
m.ruiz@berkeley.edu

**Course name/number and duration:**
Public Health 250WF: Epidemiologic Methods II: Part 1
Public Health 250WG: Epidemiologic Methods II: Part 2
Fall 2019

*R datahub or main datahub*
R datahub if possible, main datahub also works!

felder:
@ryanlovett @yuvipanda How do we go about vetting requests for admin access?  

Trust, but verify as they say.

ryanlovett:
You can [query the SIS](https://github.com/ryanlovett/sis-cli) to get a list of instructors and GSIs. A few months back [we tested a sidecar](https://github.com/berkeley-dsep-infra/datahub/commit/47d8ccb6a2c560021f1f80198ef4a1bfe70e5e97) that would do this for you automatically and deposit the lists in /srv/jupyterhub/profiles.d/. (for automatically settings admins and granting resources) I think Yuvi wanted me to use chartpress for this.

I will look into chartpress next week. The important thing is that if someone is a student in any class on a hub, they cannot also be an admin on the same hub. e.g. someone is a student in PH250WF but they're a GSI in Data 8.

felder:
For now I'm going to add the following as admins to r.datahub who are listed as instructors while we figure out a scalable procedure for adding GSIs:

ryanlovett:
Hi @anna-nguyen, we're waiting for the department that manages these course to enter the names of the GSIs into Campus Solutions. Once this happens, they will automatically become admins, unless they happen to be enrolled in another course on the same hub.

anna-nguyen:
@ryanlovett Thanks Ryan! Nolan and I aren't technically hired as GSIs, but as student assistants primarily for infrastructure support. We won't be listed as instructors on Campus Solutions - would it still be possible to get admin access?

ryanlovett:
[...]

PR #894, opened 2019-07-30
Adding jadebc and jcolford as admins.

felder:


felder:
Adding admins for #882

PR #1101, opened 2019-10-07
Get course affiliation from Canvas and assign server attributes.

ryanlovett:
Remove SIS equivalents after courses are transitioned.


ryanlovett:
@yuvipanda As discussed on slack, I haven't found a way to tell the spawner about the canvas oauth creds so for now i've set them as traitlets in our secrets. If there is a way, it can be easily changed later.

hub/values.yaml
```diff
@@ -141,8 +148,68 @@ jupyterhub:
...
+          return map(lambda x: x.get('message', ''), courses['errors'])
+
+        def course_filter(courses, enrollment_type, attr='id'):
+          '''Filter course attributes by enrollment type.'''
+          attrs = []
+          for course in courses:
+            if 'enrollments' not in course: continue
+            for enrollment in course['enrollments']:
+              if enrollment['type'] == enrollment_type:
+                attrs.append(course[attr]); break
```
yuvipanda: How do you feel about not adding multiple statements on the same line? I generally find that most python projects don't do that, and it always makes me feel like i'm reading C :)

This applies for both semicolons in stuff like this, and stuff like `             if 'enrollments' not in course: continue`

hub/values.yaml
```diff
@@ -141,8 +148,68 @@ jupyterhub:
...
+            data = {}
+            async with aiohttp.ClientSession() as session:
+              async with session.post(url, params=params) as r:
+                data = await r.json()
+            auth_state['access_token'] = data['access_token']
+            await self.user.save_auth_state(auth_state)
+            return data['access_token']
+
+          @lru_cache(maxsize=1024)
+          async def get_canvas_courses(self, token, enrollment_type=None):
```
yuvipanda: CanvasAuthenticator is also something we write customly. Can we move this to CanvasAuthenticator? We can then put the following code in CanvasAuthenticator:

1. Fetching courses
2. Dealing with expired tokens
3. Throwing errors about expired tokens

That lets us keep *just* the stuff about policy enforcment - profiles, admin determination, etc, here.

We should also move this spawner out of the YAML into the hub image, but that can wait.

hub/values.yaml
```diff
@@ -141,8 +148,68 @@ jupyterhub:
...
+            )
+            data = {}
+            async with aiohttp.ClientSession() as session:
+              async with session.post(url, params=params) as r:
+                data = await r.json()
+            auth_state['access_token'] = data['access_token']
+            await self.user.save_auth_state(auth_state)
+            return data['access_token']
+
+          @lru_cache(maxsize=1024)
```
yuvipanda: <3 the lru_cache!

hub/values.yaml
```diff
@@ -180,6 +247,59 @@ jupyterhub:
...
+            self.log.info(f"student_courses: {student_courses}")
+            self.log.info(f"instructor_courses: {instructor_courses}")
+
+            # set resource limits, add volumes, set/revoke admin based on
+            # courses
+            custom_canvas_courses = z2jh.get_config('custom.canvas_courses', {})
+            is_canvas_student = False ; is_canvas_instructor = False
+            for canvas_course_id, profile_data in custom_canvas_courses.items():
+              self.log.warning(f'evaluating course id {canvas_course_id}')
+              is_canvas_student    |= (int(canvas_course_id) in student_courses)
```
yuvipanda: In general I try to avoid bitwise operations wherever possible in Python except for actual bitwise manipulation. Can this just be:
```
is_canvas_student = int(canvas_course_id) in student_courses
```

since we only care about setting it to true at least once.

hub/values.yaml
```diff
@@ -180,6 +247,59 @@ jupyterhub:
...
+                if 'mem_limit' in profile_data and \
+                  mem_cmp(profile_data['mem_limit'], self.mem_limit) == 1:
+                    self.mem_limit = profile_data['mem_limit']
+                if 'mem_guarantee' in profile_data and \
+                  mem_cmp(profile_data['mem_guarantee'], self.mem_guarantee) == 1:
+                    self.mem_guarantee = profile_data['mem_guarantee']
+              else:
+                self.log.warning(f'user not in any canvas course')
+
+            # if the user is a student in any canvas course specified by a
```
yuvipanda: I like it! <3

hub/values.yaml
yuvipanda: You can then access the authenticator from the spawner as 'self.authenticator' (See https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/spawner.py#L150)

yuvipanda: This will also make it easier for us to upstream CanvasAuthenticator more easily.

yuvipanda: This will also make it easier for us to upstream CanvasAuthenticator more easily.

yuvipanda (changes requested):
This is awesome! Thank you so much for working on it, @ryanlovett. It's going to lay the groundwork for a lot of cool stuff :)

I've left some comments on changes I'd like to see. Let me know what you think!

hub/values.yaml
ryanlovett: Sure, can do. Apparently I like fitting code into 80x24. :)

hub/values.yaml
ryanlovett: Oh, I see, the course fetching will be invoked by the spawner, but the code will live in the authenticator?

hub/values.yaml
ryanlovett: But that would turn it off if the user wasn't a student in the next course. I can replace the bitwise operator with something more verbose though.

hub/values.yaml
yuvipanda: Yeah! I think of this code as providing 'groups' the user belongs to, and would like to push that upstream eventually. The authenticator provides this information. Various other parts can do whatever they want with this info. In our case, our spawner is doing some complex stuff to figure out who gets how much RAM...

yuvipanda:
@ryanlovett what  to do with this PR now?

ryanlovett:
Closed in favor of #2888. Thanks @yuvipanda !

PR #1156, opened 2019-11-15
Add a dependency on OpenAI Gym for the ds102 environment.yml

Karl-Krauth:
added openai gym to the dependencies.

### DIFF
--- a/deployments/data102/image/environment.yml
+++ b/deployments/data102/image/environment.yml
@@ -17,6 +17,7 @@ dependencies:
   - distributed=2.2.0
   - flake8=3.7.7
   - graphviz=2.40.1
+  - gym=0.15.4
   - h5py=2.9.0
   - hdf5=1.10.4
   - html5lib=1.0.1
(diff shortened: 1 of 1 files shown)


yuvipanda:
@Karl-Krauth heya! Thanks for opening this. Looks like the build failed for the change you added. Can you look at it? Thanks!

Karl-Krauth:
Hmm strange. Ill take a look at it deeper later today but the error message isnt giving me great diagnostics on what might be wrong (I only added gym to requirements.txt)

Karl-Krauth:
It seems that the issue is that `conda update -q conda` is somehow being trigerred in the deploy now and this results in setuptools being removed which causes a failure since conda depends on it. I'm not sure why adding a new package might cause this though.

Karl-Krauth:
Ah I see it seems that I was trying to install gym via conda when I should have been trying to do it via pip. Hopefully the tests pass now :)

PR #1160, opened 2019-11-19
scikit-image version to 0.16.2

surajrampure:


yuvipanda:
I think this has been installed in a different PR now.

PR #1178, opened 2019-12-03
Add tigris for PH142

corinne-riddell:


corinne-riddell:
@yuvipanda, this PR is to install R package `tigris`. I see it didn't pass one of the checked. Is there something I can do to make it pass?

felder:
@corinne-riddell the reason your tests are failing is tigris 0.8.2 is not tagged in the cran github repo.

You can see that here:
https://github.com/cran/tigris/tags

In order to get this to work, you can either install 0.7 from cran/tigris which is the most recently tagged, or you can use the commit id from the original tigris repo which corresponds to the existing version.

I opted to do the latter in my PR.  If you look at the commit in my linked PR above you can see how this works.  Namely, find the git repo corresponding to the package which is here:

https://github.com/walkerke/tigris and find the commit id that corresponds to the version you want.  In this case the most recent commit is the one you want.

PR #1179, opened 2019-12-03
Installing Tigris for #1178.

felder:


yuvipanda:
@felder can you help out with https://github.com/berkeley-dsep-infra/datahub/pull/1178 instead of merging this? I'd love to help @corinne-riddell be able to self-serve these requests in the future than rely on you.

Thanks.

PR #1215, opened 2020-01-15
updated requirements.txt for Astr 128/256

kareemelbadry:


yuvipanda:
@kareemelbadry can you look at the failures listed here (https://circleci.com/gh/berkeley-dsep-infra/datahub/3462?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) and try to fix them? Looks like there is a duplicate requirement listed...

kareemelbadry:
Hmm, not sure what the problem is now. The error "invalid command 'bdist_wheel'" has come up for a couple packages. A google search suggests the solution is to pip install wheel (e.g https://github.com/tensorflow/tensorflow/issues/348). But simply adding wheel to requirements.txt didn't fix the problem.

yuvipanda:
For some reason I can't update this PR (is the 'allow changes from maintainers' box checked?) so am trying to fix this in https://github.com/berkeley-dsep-infra/datahub/pull/1223

PR #1228, opened 2020-01-21
added PyPDF2 to req.txt for Rhub

### LINKED ISSUE #1222: Request python package PyPDF2 for class PH142
**Which package do you want installed?**
PyPDF2

**Which class will use this package?**
PH142

https://classes.berkeley.edu/content/2020-spring-pbhlth-142-001-lec-001

**Which semester will this package be used for?**
Spring 2020

This helps us clean up unused packages after a term ends.

**Any additional information we should know about?**

nnpok:
Closes #1222


yuvipanda:
the build failed because of lack of GitHub API limits... Going to merge and see how that goes, since the solution we have right now is to just restart builds.

PR #1237, opened 2020-01-23
added pypdf install to dockerfile for rhub

nnpok:


yuvipanda:
The issue is that the python we're installing into is `/srv/conda/bin/python` rather than `/usr/bin/python`. We should fix that...

felder:
PyPDF2 is being installed via requirements.txt.  Setting the PATH for the user environments to include /srv/conda/bin fixed I believe.  Thus I'm going to close this PR without merging.

PR #1238, opened 2020-01-23
Initial attempt to specify STAT 89A Hub

krinsman:
This hopefully doesn't fail too horribly, although I expect some syntax errors.


yuvipanda:
Hello! Thanks for opening this, @krinsman!

We support using [repo2docker](https://repo2docker.readthedocs.io/en/latest/config_files.html) config files to specify images, and I'd prefer using that over writing our own Dockerfiles where possible. In this case, I *think* we can accomplish all the things we need by specifying an `apt.txt` file and an `environment.yml` file. You can see a series of examples in https://github.com/binder-examples?page=1. This makes it easier to maintain in the long run, and you can easily test these out on mybinder.org as well.

How does that sound?

krinsman:
@ryanlovett 

1. I can look into weak versioning them. But I have in the past specified all of the exact versions in an `environment.yml`, and it actually caused more problems than it fixed. [See e.g. the Docker images I made for one project at NERSC/CRD.](https://github.com/krinsman/jupyterhub-deploy/tree/master/tomopy). Basically the problem is that if I specify the versions too much, then Conda can have too many constraints to be able to solve all of the dependencies, and then the installation fails unnecessarily. 

In any case, my short-term concern is getting a working version on staging with all of the software installed, and then I can lock in specific versions or version ranges for the rest of the semester.

2. OK, I will change it back to jovyan.

3. Again, I've done it with `environment.yml` before. [Again see e.g. the aforementioned Docker images I made for one project at NERSC/CRD.](https://github.com/krinsman/jupyterhub-deploy/tree/master/tomopy) I've found that this causes the size of the Docker images to be _substantially_ larger (even when it's specified in the YAML to install the packages in the base conda environment). 

If you don't care about the size of the Docker image being very large, then I have no problems with using `environment.yml`. I was trying to decrease the load on the system by having images that are as small as possible, because anything that's left in one layer is still part of the overall image size, even if it's deleted in a subsequent run command.

4. Again, same thing here, the idea was to attempt to minimize the image size for your benefit. ( [Again, see e.g. the aforementioned Docker images I made for one project at NERSC/CRD, where I had to choose between a multi-stage build and a long RUN statement to decrease the image size.](https://github.com/krinsman/jupyterhub-deploy/tree/master/tomopy)) Presumably I can't do a multi-stage build here. If you don't care about image size bloat, then I'll happily break up the RUN statement.

5. Again, I'm aware of this, the point of not using `environment.yml` and using a long `RUN` statement was to try to fight image bloat for your benefit. If you don't care about the size of the image, then I will happily make it as large as is convenient.

krinsman:
@yuvipanda That definitely sounds like something to explore at least for the long-term, but my short-term concern is getting a working hub setup that I can push out to all of the STAT 89A students before their lab Monday morning. I don't have any prior experience using `repo2docker`, whereas I have substantial experience writing `Dockerfile`'s from scratch for my job at NERSC/CRD, so I'm able to get something like the latter up more quickly.

krinsman:
@yuvipanda Unsurprisingly you're right, the `repo2docker` approach does have better separation of concerns and better code reuse. I wonder why DATA 100, DATA 102, and Prob 140 don't use it, but that's their loss I suppose.

Do you know any workarounds to running

```
dpkg-reconfigure --frontend-noninteractive tzdata
locale-gen
update-alternatives --install /usr/bin/bibtex bibtex /bin/true 200
```

in the Dockerfile? Or does the base Docker image in `repo2docker` already do equivalent things (set time zone to Pacific, configure locale correctly, address the NBConvert-BibTeX/LaTeX issue)? They all seem to require root to run, and as you know better than I do it's not possible to run as root in `postBuild`, and I couldn't find clear guidance from the discussion in https://github.com/jupyter/repo2docker/issues/192.

yuvipanda:
@krinsman ah, prob140 etc predate repo2docker :D that's why...

repo2docker already sets timezones and what not. And https://github.com/binder-examples/latex shows you how to do latex.

krinsman:
I rebased this to look prettier, so I'll close this PR for now.

@yuvipanda Thanks for the LaTeX pointers! Added them to the image. Pre-emptively loading the matplotlib font cache in particular is a really good idea.

PR #1245, opened 2020-01-24
Initial attempt to set up stat 89A hub

krinsman:
OK, the original intent was to merge prod into prod, but I can't change it, and apparently now I also have write access for staging, so I might as well test it on staging first anyway.


yuvipanda:
hey @krinsman. I'm going to revert this in #1247, since this is merging prod into staging which makes the git history confusing. Can you come on slack and discuss with us before doing any more merges? Thank you.

yuvipanda:
Slack would be on uctech.slack.com, in #ucb-datahubs channel. Thanks.

yuvipanda:
I'm even more confused now, because github says this PR has 250+ commits, but I don't see that in git log.

PR #1254, opened 2020-01-26
2nd attempt to set up stat 89a hub

krinsman:
Let's try again, this time without trying to muck with hub (no jupyterhub in environment.yml, no uninstall pip jupyterhub, etc etc)


krinsman:
It actually worked this time! Thank goodness

PR #1312, opened 2020-02-07
Lower guarantee for data100 to 512

yuvipanda:
Followup to https://github.com/berkeley-dsep-infra/datahub/pull/1279
as no issues were reported


yuvipanda:
@wwhuang I'll merge this on Monday

PR #1314, opened 2020-02-07
Add a storage retention policy

yuvipanda:


yuvipanda:
@ryanlovett has written https://github.com/berkeley-dsep-infra/share-user-data to do this.

ryanlovett:
Originally by @simon-mo.

papajohn:
Seems like a good plan, but I would have thought 1 year would be the magic time to wait before archiving (instead of 6 months), but I don't feel strongly either way. I'm glad the data is being preserved. 

Some guide for how to move the data back to datahub will be requested by some student at some point, so we could just write it now.

rameshvs:
I agree with John: I think one year would be better than 6 months (since that allows students to see material from 2 semesters ago). 

Also, it would be nice to have a semi-automated process for students to follow to restore their data: at the minimum, instructions on how to upload the backed-up data from google drive back to datahub, and in an ideal world, a button they can click to automatically have their data restored.

davidwagner:
Looks good to me!  6 months seems generous to me.  I had thought we didn't make any promises after the end of the semester, and I previously had in my mind that students should assume their data might be deleted shortly after the end of each semester.  Even 2 months seems like it would be adequate, from my perspective.  Given the resourcing challenges we face at Berkeley, I think any effort at all to archive data and make it available to students is icing on the cake.  Just my two cents.

papajohn:
I do feel strongly that students should have an easy path to accessing a
computing environment they understand if they want to mess around later.
That’s one reason why the hub is available to all Berkeley affiliates, not
just current students.

yuvipanda:
I've upped it to 12 months. I mostly wanna make sure we have *some* policy in place, and it's communicated clearly.

We can write some docs on how students can download the zip file and re-upload it.

yuvipanda:
I'm going to merge this on Feb 17, Monday if there are no more objections

yuvipanda:
Turns out Feb 17 was a holiday, so I'll merge it Feb 19 (Wednesday). So one more day!

yuvipanda:
I hereby deem this our storage retention policy!

PR #1333, opened 2020-02-14
alphabetize pip packages

krinsman:
not sure this will bump version of linalg_for_datasci when merged into prod, but worth a shot


yuvipanda:
Nope, pip doesn't care about this order https://stackoverflow.com/a/11597690

PR #1352, opened 2020-03-04
Add 4G memory limit to astro 128 students

yuvipanda:


deployments/datahub/config/common.yaml
```diff
@@ -28,6 +28,10 @@ jupyterhub:
         2019-fall-31002: {}
         # data 88
         2019-fall-33469: {}
+        # astr 128/256
+        2020-spring-astron-128-001-lab-001:
```
ryanlovett: @yuvipanda it should actually be `2020-spring-24268`, based on the "Class #" in the listing.

yuvipanda: @ryanlovett aaah, thanks. How about now?

ryanlovett: Looks good!

PR #1407, opened 2020-04-22
Revert #1404 due to gofer test issues

chrispyles:
Reverts berkeley-dsep-infra/datahub#1404


chrispyles:
Closing so that I don't revert #1406

Issue #1456, opened 2020-05-12
Use miniforge to get Python


yuvipanda:
We currently use many different methods to install Python:

1. System python, in datahub, data8x
2. [miniconda](https://docs.conda.io/en/latest/miniconda.html) directly, for data102, data100 and r
3. repo2docker provided miniconda, for julia, desktop-test, stat89a

We should standardize on [miniforge](https://github.com/conda-forge/miniforge/) as the way to get python instead.

1. It is fully community maintained, unlike miniconda
2. It pulls from the community maintained [conda-forge](https://conda-forge.org/), which is what we want to support
3. We can unhook python version from system python version. Without that, upgrading version of Python requires upgrading base image.

mybinder.org has been running this for a while with no ill effects, so it is time to move. We will generally continue using just pip to get our dependencies, since I don't want us to mix multiple sources for package installs.

This should be done in stages.

- [ ] Use a newer version of repo2docker with [https://github.com/jupyter/repo2docker/pull/859](this PR I made) so they will automatically switch to miniforge.
  - [ ] julia
  - [ ] desktop-test - #1457
  - [ ] stat89a
- [ ] Move r hub over, since it doesn't really use an environment.yml file
- [ ] Move data8x over, but with the exact same version of Python. This lets us test the effects of the change without it being too disruptive.
- [ ] Move data100. Don't touch the environment.yml file, don't change sources. See how that goes.
- [ ] Do the same for data102. 
- [x] Move datahub, but keep python version fixed

Semester ends may 15th, so I will try to do most of these things after.

ryanlovett:
I'd like to move prob140 to repo2docker as well. Once the semester is over I'll submit a PR with [prepared changes](https://github.com/berkeley-dsep-infra/datahub/compare/staging...ryanlovett:prob140-2020-fall?expand=1).

yuvipanda:
@ryanlovett awesome! I opened it as a PR anyway...

ryanlovett:
Okay, didn't want to block, but no problem that its a PR now. Looks like it failed CI -- will iterate.

yuvipanda:
Done! |o/

PR #1457, opened 2020-05-12
Use newer version of repo2docker

yuvipanda:
This uses miniforge instead of miniconda to get python.

See https://github.com/berkeley-dsep-infra/datahub/issues/1456
for rationale.

We make a small change in desktop-test to trigger a
rebuild, so it can be rebuilt to use miniforge

Ref #1456


yuvipanda:
Gonna wait till next week before merging this.

PR #1499, opened 2020-06-08
Move data100 to using repo2docker

yuvipanda:
There isn't any strong reason to use a Dockerfile here.
repo2docker makes the image a little bigger, but that's
totally ok in our case.

The update-alternatives for bibtex isn't present in
the Dockerfile, but hopefully that should be ok.


ryanlovett:
> The update-alternatives for bibtex isn't present in the Dockerfile, but hopefully that should be ok.

@yuvipanda That was an old workaround for TeX-related bugs in nbconvert. I think the hubs will soon standardize on chromium-based notebook conversion anyways.

yuvipanda:
@ryanlovett right i vaguely remember. Do you know how to test for that?

ryanlovett:
@yuvipanda it has [been a while](https://github.com/berkeley-dsep-infra/data100-19s/commit/dcf5d20094ddfd448b1318a90df2f6e7a4f9cbd1), but if I recall the save-to-pdf was broken for many of the notebooks. The [Data 100 materials repo](https://github.com/DS-100/sp19) for [Spring '19](http://www.ds100.org/sp19/syllabus) doesn't exist anymore, but I suppose you could try to reproduce with sp20 notebooks.

For context, I just forwarded the email thread which prompted the bibtex override.

PR #1535, opened 2020-06-16
Add notetaker to Dockerfile for Data 100 hub study

chrispyles:
This PR updates the Data 100 default Docker image to install and enable the nbextension https://github.com/yifanwu/notetaker after cloning from git to `/tmp/notetaker`.


chrispyles:
@yuvipanda has advised that this needs to be a Python package. Closing this PR so that that change can be made.

Issue #1574, opened 2020-06-25
Use bcourses auth for all hubs


yuvipanda:
Currently, not all hubs have bcourses auth - some still use Google Auth via Calnet. We should move it all to bcourses.

To move
- [x] data100.datahub.berkeley.edu
- [x] data102.datahub.berkeley.edu
- [x] stat89a.datahub.berkeley.edu
- [x] eecs.datahub.berkeley.edu
- [x] prob140.datahub.berkeley.edu

Issue #1453, opened 2020-05-10
Upgrade version of notebook package across all hubs


yuvipanda:
The `notebook` python package is at 6.x now. We have been running a full version behind (5.x) for a while now. Time to upgrade!

We should do this in a staggered fashion, one hub at a time.

- [x] prob140
- [x] r
- [ ] datahub
- [ ] julia
- [ ] data8x
- [ ] workshop
- [x] data102
- [ ] stat89a
- [ ] template
- [ ] desktop-test

felder:
r hub upgraded

felder:
data102 is running notebook 6.0.0

felder:
stat89a doesn't build if I specify a new notebook version in environment.yml

yuvipanda:
This is now done, with #1667

PR #1604, opened 2020-07-07
Upgrading notebook and adding bcourses auth.

felder:


felder:
Issue #1453 and #1574

PR #1620, opened 2020-07-11
Update common.yaml with EECS16A admin users

Quizenger:


deployments/eecs/config/common.yaml
```diff
@@ -31,6 +31,25 @@ jupyterhub:
           - yuvipanda
           - felder
           # List of other admin users
+          - lakabuli
```
yuvipanda, line 34: Can you add a comment here including which class these users are added for, and when they can be removed? That will help us keep this clean. Thanks!

Quizenger: For sure. Could I change the semester when they could be removed later? We expect to use Datahub for Fall 2020 and potentially Spring 2021. I could update the list with accounts to be removed once they leave course staff.

yuvipanda: @Quizenger yep, you can! Ideally at beginning of each semester the list would be changed to match reality.

PR #1657, opened 2020-08-04
Remove 'rebalancer'

yuvipanda:
We haven't actually used this in ages, since it wasn't
good enough to work with multiple hubs. Now, gone for good.

New nodes taking too long to start up was the problem it was
trying to solve. We have been able to do better there by splitting
our nodes into separate nodepools, so fewer images need to be
pulled per-node. That seems to have worked well enough for now.


yuvipanda:
This was added in e181e1a64f741d3a44cf2a02b900b15b7b8a4654, the commit message provides good rationale.

Issue #1461, opened 2020-05-13
Provide clear way to download notebook as PDF


yuvipanda:
The following download options are available for students:

<img width="403" alt="image" src="https://user-images.githubusercontent.com/30430/81810598-d871be80-9540-11ea-8f58-7673335d39b5.png">

This is way too many, and most of them are useless.

Instead, we should just have the following options:

- Notebook (.ipynb)
- Python (.py)
- PDF

Currently PDF requires LaTeX. I don't want us to use LaTeX. 

https://github.com/betatim/notebook-as-pdf is probably the best current way to generate PDFs from notebook files. We currently have [my version](https://github.com/yuvipanda/nbpdfexport) installed in a few hubs. We should standardize on this instead.

For other formats, we can offer the appropriate nbconvert command so folks can run and download it manually

yuvipanda:
This will make @ericvd-ucb very happy

yuvipanda:
Asked question in https://discourse.jupyter.org/t/remove-download-as-options-in-jupyter-notebook/4374

ericvd-ucb:
Yes! - it will make me happy - and many confused newbies as well.
I guess I don't exactly get what the betatim workflow does , what "attached" means, and what the limitation of not opening in preview and have to have acrobat ( FWIW I have preview but not Acrobat) 

@ktakimoto  or  @chrispyles  Want to weigh in here?

yuvipanda:
@ericvd-ucb i think @betatim's version should work exactly the same as mine, except it also lets you retrieve the original notebook.ipynb file from the PDF if you have the appropriate software. The PDF itself should open in preview or other PDF applications

ktakimoto:
Seconding the happiness around cleaning up the pdf options!

@ericvd-ucb having the ipynb file attached to the pdf is intriguing as we think about uploading assignments to Gradescope. I think our initial idea was to ask student to upload the ipynb, then convert to pdf on Gradescope- with this version, could we do the reverse (students upload pdf, ipynb is extracted on Gradescope)?

re: preview question, I think our OSX users don't need to see the ipynb attachment as long as they can see the pdf (and don't get any weird error messages about hidden attachments- can't test this currently, as I'm on Windows).

[...]

PR #1691, opened 2020-08-18
Replace nbpdfexport with notebook-as-pdf

yuvipanda:
Is actually maintained!

Ref #1461


yuvipanda:
@chrispyles said it wasn't being particularly used by anyone.

PR #1704, opened 2020-08-19
data100: Bump to python 3.8

yuvipanda:
Let's try this out


yuvipanda:
@chrispyles objections? I'll reach out to the data100 instructors too

chrispyles:
@yuvipanda I have no objections but yes would recommend confirming with data 100 instructors as well

Issue #1699, opened 2020-08-19
Request PAUP* for Bio 1B course on Datahub


Elias326:
**Which package do you want installed?** PAUP*

**Which hub do you want it installed on?  For example r.datahub, datahub, data100.datahub, etc.** Datahub

**Which class will use this package?** Bio 1B

Include a link to appropriate entry in https://classes.berkeley.edu/ if available. If not, please mention class name & instructor.

**Which semester will this package be used for?** Fall 2020 / Spring 2021

This helps us clean up unused packages after a term ends.

**Any additional information we should know about?**

yuvipanda:
Thanks for opening this request, @Elias326!

Can you point me to where I can find more information about this software? I've only found https://paup.phylosolutions.com/get-paup/, but it looks like a commercial solution you have to pay for if you want to use it in a non-trial fashion? We generally try to avoid those on our hubs, since that makes management more difficult. Does the bio department have a license for it or somesuch?

ericvd-ucb:
Hi @yuvipanda 
I have looped back with the instructor and he really wants to try this out - and really wants to get it on Datahub
He is basically trying to make phylogenetic trees using dendropy from using real data

on this page http://phylosolutions.com/paup-test/  it claims that the CLI version will be FOSS but the GUI version will be for pay.  But yes he claims it would be licensed freely by the profs who wrote it for instructional use

I am going to forward you a couple email threads

yuvipanda:
I think this is done.

PR #1717, opened 2020-08-21
biology: Rename genomics to biology

yuvipanda:
With #1699, we now have more than one biology class
on datahub. Makes sense to move genomics into something
more general


petersudmant:
@yuvipanda - if I want to convert this to a jupyterlab (for IB134L - aka "genomics"), will that create problems for BIO1B?

PR #1752, opened 2020-08-29
Revert "Prod"

yuvipanda:
Reverts berkeley-dsep-infra/datahub#1749

Generally, only staging should be merged to prod, otherwise we get all these
strange merge commits into staging.


yuvipanda:
Except this doesn't seem to work either...

PR #1788, opened 2020-09-03
Upgrade to R 4.0.2.

ryanlovett:


ryanlovett:
The "package" package doesn't address the RStudio failures here.

PR #1803, opened 2020-09-07
Add requirements for ds198 mch infodemiology

azepecki:
I am a guest instructor for DS198 MCH Infodemiology and am coordinating the students' technical setup from the Wallace Center team side. 

These requirements are necessary for the Jupyter notebooks we will be distributing to the students.

### DIFF
--- a/deployments/datahub/images/default/requirements.txt
+++ b/deployments/datahub/images/default/requirements.txt
@@ -140,3 +140,10 @@ lcapy==0.66.0
 
 # EPS 256, https://github.com/berkeley-dsep-infra/datahub/issues/1775
 obspy==1.2.2
+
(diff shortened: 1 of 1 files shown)


yuvipanda:
Thank you for making the PR, @azepecki! Makes everything much easier :)

azepecki:
@yuvipanda Is there any delay in when these packages will be available? I am still not able to load the libraries from the required packages I added.

yuvipanda:
@azepecki apologies, I forgot to merge https://github.com/berkeley-dsep-infra/datahub/pull/1804. It should be available in about 20-30 min now.

azepecki:
Thanks! All good to go @yuvipanda 👍

Issue #653, opened 2019-04-27
Request to have two packages installed on the dataHub


mdshapiroLBL:
I would like to request two packages be added to the data hub:  ROOT and pythia8.  These packages are both widely used by particle physicists and have well tested python interfaces that function well through Jupyter.  We plan to use these packages for Jupyter based problems sets in two Physics classes (Physics 129 and Physics 226).

Both packages are available for installation via conda.   The relevant information is available at the sites listed below:

ROOT:  [https://github.com/conda-forge/root-feedstock](url)

pythia8:  [https://github.com/conda-forge/pythia8-feedstock](url)

These packages are also available in docker containers, but if the conda installation works, that would be a preferable solution.

Please let me know if you have any questions about these packages or any issues with installation.

Thanks,
Prof Marjorie Shapiro
Physics Department

yuvipanda:
Thanks for making the requrest here, @mdshapiroLBL!

Do you know if these packages are also available from pypi.org / installable via pip? On datahub.berkeley.org we use pip right now, rather than conda...

Also can you tell us what the timeline for the classes is? 

Thanks.

mdshapiroLBL:
Thanks for the rapid response.  We are targeting Fall 2019 for the classes.

I don't think that we can use pip to install.  Both ROOT and pythia8 are
written in C++ with python bindings.  I didn't think pip could handle that
(although I would be happy to be told that I am wrong).   Conda uses CMake
to handle the various OS dependencies.

BTW, I tried to see what version of gcc is native for the dataHub and was
surprised to find that when I do:

import os
os.system("gcc -v")

From my dataHub account I get:

ModuleNotFoundError: No module named 'gcc'

I have never seen a Linux system without gcc as part of the environment, so
I am a bit confused.

ROOT is the more important of the two packages for us, so let's start with
that one. If you think using Docker is an easier way to access this
package, here is a pointer to the relevant documentation for ROOT:
https://hub.docker.com/r/rootproject/root-ubuntu16/

ryanlovett:
CERN provides some binaries for ROOT at https://root.cern/content/release-61206 -- latest supported Ubuntu is 17.x.

yuvipanda:
@mdshapiroLBL when I run 'gcc -v' in a terminal ('New -> Terminal' from the Jupyter Notebook directory listing page), it tells me we've gcc 7.3 installed. Not sure why os.system isn't finding it.

[...]

PR #1811, opened 2020-09-08
Installing root and pythia8 via conda for #653

felder:


felder:
@yuvipanda something is really wrong with conda, gcc, and any module that requires compilation.  Namely it's not looking for libraries or headers in the correct places.

If you go through the build logs as well as my commits in this PR you can see me playing whack-a-mole with packages, but this is just not going to work.

My best guess is that this is similar to:
https://github.com/ContinuumIO/anaconda-issues/issues/11152

For example:
gcc -pthread -shared -B /opt/conda/compiler_compat -L/opt/conda/lib -Wl,-rpath=/opt/conda/lib -Wl,--no-as-needed -Wl,--sysroot=/ build/temp.linux-x86_64-3.8/_commander.o -lm -o /opt/conda/assignment-1910/cosmofast/planck2018/_commander.cpython-38-x86_64-linux-gnu.so

/opt/conda/compiler_compat/ld: cannot find /lib64/libpthread.so.0

/opt/conda/compiler_compat/ld: cannot find /usr/lib64/libpthread_nonshared.a

collect2: error: ld returned 1 exit status

error: command 'gcc' failed with exit status 1

PR #1865, opened 2020-10-02
add R studio support to Biology

petersudmant:
added R studio support and moved hub to use
Dockerfile based on Dockerfile here:
https://github.com/utoronto-2i2c/jupyterhub-deploy/tree/staging/deployments/utoronto/image


jupyter_notebook_config.py
```diff
@@ -0,0 +1,12 @@
+# This is sh (dash) by default, not $SHELL
+c.NotebookApp.terminado_settings = { "shell_command": ["bash"] }
```
yuvipanda, line 2: This file was accidentally added?

deployments/biology/image/jupyter_notebook_config.py
```diff
@@ -0,0 +1,12 @@
+# This is sh (dash) by default, not $SHELL
+c.NotebookApp.terminado_settings = { "shell_command": ["bash"] }
+
+c.ServerProxy.servers = {
```
yuvipanda, line 4: Let's remove this config setup, don't think we need this here

deployments/biology/image/install.R
```diff
@@ -0,0 +1,47 @@
...
+
+# Install devtools, so we can install versioned packages
+install.packages("devtools")
+
+# Install a bunch of R packages
+# This doesn't do any dependency resolution or anything,
+# so refer to `installed.packages()` for authoritative list
+cran_packages <- c(
+  "tidyverse", "1.3.0"
+  #"learnr", "0.10.1",
```
yuvipanda, line 11: Kill all commented out packages here?

deployments/biology/image/environment.yml
```diff
@@ -5,13 +5,16 @@ dependencies:
 - python=3.8.*
 - pip=20.*
 
+# nodejs
+- nodejs
```
yuvipanda, line 9: For R, we should get nodejs from ubuntu - that's what some of the R packages that use js want (particularly, `rstan`). This gives us an older nodejs but that's fine.

deployments/biology/image/Dockerfile
```diff
@@ -0,0 +1,189 @@
...
+
+# Install PAUP* for BIO 1B
+# https://github.com/berkeley-dsep-infra/datahub/issues/1699
+RUN wget http://phylosolutions.com/paup-test/paup4a168_ubuntu64.gz -O ${CONDA_DIR}/bin/paup.gz
+RUN gunzip ${CONDA_DIR}/bin/paup.gz
+RUN chmod +x ${CONDA_DIR}/bin/paup
+
+############################
+# Install packages for IB134L
+############################
```
yuvipanda, line 166: Let's put this into its own bash file that can be called from here?

deployments/biology/image/Dockerfile
```diff
@@ -0,0 +1,189 @@
...
+# Our pre-built R packages from rspm are built against system libs in focal
+# rstan takes forever to compile from source, and needs libnodejs
+# So we install older (10.x) nodejs from apt rather than newer from conda
+RUN apt-get update -qq --yes > /dev/null && \
+    apt-get install --yes -qq \
+    r-base \
+    r-base-dev \
+    r-recommended \
+    r-cran-littler \
+    nodejs
```
yuvipanda, line 54: We need to add 'npm' here, along with `nodejs`. Otherwise jupyterlab extensions will fail to build

yuvipanda (commented):
Thanks a lot for doing this! I've left some comments on improving this image.

PR #1873, opened 2020-10-05
Give the hub container time to startup

consideRatio:
The default value for the livenessProbe which is already 60 seconds, so
what's needed is to have an even longer duration than 60 seconds.

Note that the readinessProbe only influence the Ready state of the hub container, which in turn infleunce the k8s Service resource that will only direct traffic to pods with all their containers in a Ready status.

This is a followup to https://github.com/berkeley-dsep-infra/datahub/pull/1871.

### DIFF
--- a/hub/values.yaml
+++ b/hub/values.yaml
@@ -110,9 +110,9 @@ jupyterhub:
             - namespaceSelector:
                 matchLabels:
                   name: support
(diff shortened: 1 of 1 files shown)


yuvipanda:
Thanks a lot, @consideRatio!

PR #1879, opened 2020-10-06
Tune NFS client mount options

yuvipanda:
We were getting latency spikes (upto 5 min sometimes!)
on notebook saves. Trying various options, including this.


consideRatio:
Wieee thanks for sharing this with me @yuvipanda! You work on NFS stuff is... 😍

Issue #1899, opened 2020-10-09
RStudio Initialization Error - Discussion and Workaround


cdbeon:
Hello all,

I've had a few students come up to me with the following error when trying to log into their RStudios:

`RStudio Initialization Error: Error occurred during transmission.`

The first report that I got of this issue came early today at around 12 AM.

Restarting their servers (going to the admin hub, clicking `stop server`, then `start server`) didn't seem to work either.

Any suggestions as to how to resolve this issue?

Thank you!

felder:
This same issue was also reported yesterday on r.hub on behalf of another student by @d-alex-hughes

fdrennan:
I'm here because I'm having a similar issue on RStudio Server. 

Here's what I know. When I am using the preview version, I get the error. When I revert to the current release of RStudio Server, the error disappears. When reinstalling RStudio Server Preview (or daily build) the error returns. I am able to get to the log in but the error occurs after submitting my password and before the IDE appears.

felder:
@lytello @d-alex-hughes @cdbeon Any other students report this issue?  If not I may close this for now and reopen later if this appears to be a consistently occurring problem.

d-alex-hughes:
Thanks for digging into this with the student who raised it @felder. 🎉  Their instance is running sans problems now and I haven't had any new cases bubble up in the last few days. 

If it does rise up again, I'll bring it back into this issue and notify you.

lytello:
@felder I had 5 more cases (likely more that didn't require my assistance). The ones that did reach out had lost their work. From my asking questions, these students weren't regularly saving their files when using datahub, which I suspect is why this additional issue occurred.

blulightspecial:
Just had another MIDS student raise this issue on slack.

ericvd-ucb:
Someone just reported this via DS-infrastructure email as well.

blulightspecial:
@ericvd-ucb it is likely that MIDS student. I directed them to reach out via email.

cdbeon:
@felder I've had 3 more cases since then

felder:
@cdbeon any word on how to reproduce?

cdbeon:
[...]

PR #1922, opened 2020-10-26
Delete rstudio state file on startup

yuvipanda:
Seems to be causing a bunch of blocking issues for students,
where they are unable to access the R hub because of possible
corruption in RStudio. Deleting / renaming the state file
seems to do the trick, but doing that on a per-student basis -
especially if they are blocked from using the hub in the meantime -
seems bad.

This might affect classes that rely on persistent R state,
but in general I think that's not a recommended pattern -
preferring state to be in explicitly saved files &
code itself (I could be totally wrong!). We can merge this
to unblock users, and see if others complain about the lack
of state?

Ref #1899

----

#

### DIFF
--- a/deployments/r/config/common.yaml
+++ b/deployments/r/config/common.yaml
@@ -48,6 +48,18 @@ jupyterhub:
           # List of other admin users
 
   singleuser:
+    initContainers:
+      # Delete rstudio session on startup
+      # See https://github.com/berkeley-dsep-infra/datahub/issues/1899
+      - name: delete-r-session
+        image: busybox
+        command: ["sh", "-c", "rm -rf /home/jovyan/.rstudio"]
+        securityContext:
+          runAsUser: 1000
+        volumeMounts:
+        - name: home
+          mountPath: /home/jovyan
+          subPath: "{username}"
     nodeSelector:
       hub.jupyter.org/pool-name: gamma-pool
     storage:


yuvipanda:
Here's a small sample of sizes of `.rstudio`:

```
380K
134M
64M
52K
8.7M
4.7M
185M
348M
140K
52K
1.1M
56K
236K
201M
15M
180M
82M
61M
52K
18M
49M
286M
176K
6.4M
178M
188M
52K
237M
3.6M
145M
1.6M
22M
836K
96K
257M
251M
177M
52K
96K
80K
412M
1.1M
564M
88M
8.8M
1.9M
39M
175M
224M
558M
13M
74M
```

Not all of them are super small, unfortunately.

d-alex-hughes:
Removing '~/.rstudio/` seems like a nice, minimal solution to solving this.

To my understanding [this post](https://support.rstudio.com/hc/en-us/articles/218730228-Resetting-a-user-s-state-on-RStudio-Server) from Rstudio support says this proposed change will remove active and suspended sessions, but will maintain user settings and open code files. 

Location | When deleted
-- | --
~/.config/rstudio/ | All user settings (Global Options) lost
~/.local/share/rstudio/ | All open files and tabs lost
~/.rstudio/ | All active and suspended sessions deleted

d-alex-hughes:
If possible, can we document this behavior? It is different from the out-of-the-box behavior for student's local installs and may catch some off-guard the first time.

yuvipanda:
@d-alex-hughes I'm going to close this in favor of https://github.com/berkeley-dsep-infra/datahub/pull/2035. That should give users on datahub a different .rstudio from r hub, and *hopefull* solves some of these issues. New versions of R & RStudio are also coming!

PR #1975, opened 2020-11-18
Self-shutdown notebooks after 1h of no activity

yuvipanda:
The culler doesn't kill some user pods when it should. We now
get the notebook server to try and stop itself when it's got
no activity either. Not sure if this will help, but let's
try! This is what mybinder.org does.

See https://github.com/jupyterhub/jupyterhub/issues/3099
for more info.

### DIFF
--- a/hub/values.yaml
+++ b/hub/values.yaml
@@ -20,7 +20,15 @@ etcJupyter:
       # https://github.com/ipython/ipython/issues/9163
       db_file: ":memory:"
     NotebookApp:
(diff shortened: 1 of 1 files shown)


yashali:
@yuvipanda Did this help you with your issue on https://github.com/jupyterhub/jupyterhub/issues/3099 ?

This setting doesnt work as expected at all at our end. Does the jLab browser tab have to be closed for this to work? I had opened Jlab on one tab, switched to another tab  and minimized the browser to test this config

Issue #2002, opened 2020-12-07
Spring 2021 cleanup tasks


yuvipanda:
List of things to do before Spring 2021 classes start.

- [ ] Update all packages & libraries to newer versions. Particularly, move all base dockerfiles to Ubuntu 20.04 from 18.04
- [x] Decide on an R / RStudio story for datahub, and implement it
- [x] Roll out home page to all hubs
- [ ] Archive unused home directories of users, per archival policy
- [ ] Setup per-hub XFS quotas, so we don't accidentally overrun them
- [x] Do a round of cost optimizations
- [x] Re-evaluate XFS mount options (perhaps with noatime)

Optional:

- [ ] Try determine metrics for how much the admin dashboard is used
- [ ] Email folks who are using the hubs and check in on their experiences

ryanlovett:
How about removing the dependencies of old courses?

yuvipanda:
For R, I have a slightly different proposal.

1. Integrate current datahub & R hub images, making sure everything in the R image is in the base datahub image.
2. Use the same image for both hubs.

This should help reduce the maintenance burden of two separate images. It will also support users who mix R & python libraries in their code a lot, a [common pattern](https://rstudio.github.io/reticulate/).

Once #2008 is in place, we can consider redirecting the R hub to datahub. But it's probably still very useful in terms of social signaling, so we should keep it for now.

yuvipanda:
Based on https://utcc.utoronto.ca/~cks/space/blog/linux/NFSReadonlyAtime, I've now actually remounted our XFS mounts to have `noatime,nodiratime` explicitly.

```
yuvipanda@nfsserver-01:~$ mount | grep noatime
/dev/sdb on /export/homedirs-other-2020-07-29 type xfs (rw,noatime,nodiratime,attr2,inode64,logbufs=8,logbsize=32k,prjquota)
/dev/sdd on /export/datahubhomes-2020-07-29 type xfs (rw,noatime,nodiratime,attr2,inode64,logbufs=8,logbsize=32k,prjquota)
/dev/sdc on /export/data100homes-2020-08-04 type xfs (rw,noatime,nodiratime,attr2,inode64,logbufs=8,logbsize=32k,prjquota)
```

`sudo mount -o remount` worked fine. Let's see how this affects our stats in http://grafana.datahub.berkeley.edu/d/j2uCW9vGz/nfs-stats?orgId=1&refresh=30s

ryanlovett:
[...]

PR #2028, opened 2021-01-04
Upgrade datahub & data100 images to focal

yuvipanda:
- New versions of all apt packages!
- R can now install binary packages from packagemanager.rstudio.com
  This speeds up everything. We also get rid of R package installs
  from apt.
- Newer version of R & RStudio
- Latest versions of all R packages mentioned here are picked
  from packagemanager.rstudio.com
- No python or python libraries version bump.
- data100 also bumped.

Ref #2002


yuvipanda:
The datahub image builds fine on my local machine, so I'll chalk this to weird CircleCI behavior.

PR #2034, opened 2021-01-08
datahub: Use newer version of R

yuvipanda:
focal still ships with 3.6 only


yuvipanda:
Another case of image building successfully locally, but failing on Circle CI. I ran `hubploy build datahub --push`, so the image has been pushed. CircleCI needs fixing though.

PR #2072, opened 2021-01-15
Add data100 packages to datahub image

yuvipanda:
Let's move data100 to the datahub image. data100 hub is
primarily to limit admin access. Using the same image reduces
our maintenance burden. We couldn't do this early on because
the datahub image did not use conda. It does now.

Installs the jupyter_nbextensions_configurator on all hubs
as well, since it was in the data100 hub. Seems generally
useful.


yuvipanda:
I can't get jupyter_nbextensions_contrib to work yet, but will do that in another PR

PR #2096, opened 2021-01-19
Install R packages as NB_USER.

ryanlovett:


yuvipanda:
Based on https://packagemanager.rstudio.com/client/#/repos/1/packages/units, this might need `apt-get install -y libudunits2-dev`

ryanlovett:
@yuvipanda Thanks! It's weird that it didn't miss the dependency when installing as root.

Issue #2062, opened 2021-01-14
Update new-hub.rst and cookiecutter template


ryanlovett:
I deployed stat159-staging today by following the new-hub docs and came across some items there and in the cookiecutter template that I think should be updated. I'll patch everything if there's no other comments:

1. The cookiecutter template suggests "dns-valid-hub-name" as a value for "hub_name". It really means the short hostname (the name before the first dot), and not the fqdn or "dns-valid". I recommend that the hint be changed to "dns-hub-short-name".
2. The template needs to change nfsPVC.nfs.serverIP to `nfsserver-01`.
3. Between steps 4 and 5, you have to 1) add the new deployment to `.circleci/config.yml`, and 2) create new k8s namespaces manually. This used to be done by helm2.
4. `secrets/gcr-key.json` and `secrets/gke-key.json` are not part of the cookiecutter, so they need to be copied over from one of the other deployments.
5. Step 7 should refer to config/staging.yaml, not secrets/staging.yaml
6. The template needs a new jupyterhub-ssh.hostKey variable in each secrets file with `ssh-keygen` hint, along with jupyterhub-ssh.hubUrl in the staging/prod config files.
7. Step 9 needs to be done before step 5 due to jupyterhub-ssh
8. Add doc about canvas oauth client id and secrets. From @felder, "so datahub has two developer keys in canvas. one is for all of the staging hubs (they all share a key). one is for the prod hubs (they all share a key)". we should ask @felder to ask Miles to update the hub list in bcourses. If @felder is not available, we should email bcourseshelp@berkeley.edu with a specific mention of Miles or a cc to him.
9. Document adding new netreg offsite hosting entries for prod/staging in the "CDSS, RTL Supported" security context.

felder:
@ryanlovett Miles is ok with me updating the list myself.  So generally speaking, just ask me and I'll make the change.  In the event I'm not available, then send the email to bcourseshelp@berkeley.edu

felder:
I too found the cookiecutter hint "dns-valid-hub-name" to be confusing.  I think adding something like "dns-valid-short-name (e.g. datahub, r, or data100)" would be helpful here.

felder:
cookiecutter template needs to set images.images.path and images.images.name in hubploy.yaml

felder:
jupyterhub.auth.type needs to be changed from google to custom to use bcourses auth

yuvipanda:
This has been updated now!

PR #2115, opened 2021-01-25
Amend new-hub infrastructure docs

yuvipanda:
Ref #2062


felder:
@yuvipanda Looking over the docs:

1. jupyterhub.hub.cookieSecret is missing entirely.

2. Also it's unclear from the documentation whether or not the same values are placed in the secrets file for both production and staging or if they require different ones.  The docs appear to indicate prod and staging should be done as two separate procedures, but then the docs ALSO tell you to fill in values for staging and prod at the same time (step 5 and step 13). I think the docs would be clearer if they indicated which tasks have to be performed once (I assume cookiecutter is once) and which ones need to be performed separately for staging and prod.

3. Step 7 is missing the actual example stanza.

4. Link in step 9 is broken.

felder:
@yuvipanda 

I think I'm going to surrender on attempting to create dlab otherwise I'm going to have to bug Ryan throughout the afternoon.

IMO, the current docs are not sufficient to allow someone who has not done this before to succeed.

yuvipanda:
Thank you for all this feedback, @felder. Sorry it was a frustrating experience :( 

I've updated this PR with new docs based on your feedback here and on slack. I've also automated *all* secret generation, so the hub setup process is much simpler!

Issue #2116, opened 2021-01-25
New hub for cs194


yuvipanda:
The [data engineering class](https://www2.eecs.berkeley.edu/Courses/CS194_4349/) by Joe Hellerstein and Aditya Parameswaran requires a postgres database per user that's persistent. We did the postgres stuff for data100 last time, but it isn't persistent.

So we'll:

- [x] Add OAuth callback entries to bcourses for two hubs: https://cs194-staging.datahub.berkeley.edu and https://cs194.datahub.berkeley.edu
- [x] Create a new hub (cs194.datahub.berkeley.edu)
- [x] Copy the postgres sidecar config to it from data100
- [x] Give each user a 1GB persistent disk
- [x] Put the database on this disk
- [ ] Allow access *only* to users enrolled in the class

felder:
@yuvipanda Added the callback entries to bcourses.

PR #2117, opened 2021-01-25
Use newer postgres image for data100

yuvipanda:
Also fix extraImages, which was pulling *wrong image*

Ref #2116


yuvipanda:
I ran `chartpress --push` manually to build the image.

PR #2124, opened 2021-01-26
cs194: Provide persistent storage for postgres

yuvipanda:
For each user, we create a PVC and mount the postgres
database directory there. We also mount the same directory
in the user pod, in case direct inspection is necessary.

Each user gets 1G ssd. For about 200 users, that's 200G ssd -
which is as much each node gets. I think we're ok cost-wise for now.

Ref #2116

### DIFF
--- a/deployments/cs194/config/common.yaml
+++ b/deployments/cs194/config/common.yaml
@@ -14,6 +14,49 @@ jupyterhub:
       hub.jupyter.org/pool-name: core-pool
 
   hub:
+    extraConfig:
+      cs194-01-pvc: |
+        storage_class = 'ssd'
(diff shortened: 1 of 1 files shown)


rabernat:
Just wanted to chime in and say I love this sort of feature. It would be great to be able to play with all sorts of persistent databases on our hubs. I personally would use Mongo and Redis eagerly. Those are also offered from cloud providers directly.

yuvipanda:
great to hear that, @rabernat! This gives each user their own postgresql instance. Do you imagine that's the kinda stuff that folks would wanna do with mongodb or redis? It is very persistent though, so might help

PR #2152, opened 2021-02-04
Pin nbformat<5.1.0

yuvipanda:
Per @chrispyles


yuvipanda:
@chrispyles do you have a link to notebooks that might have issues with this? If this is https://github.com/jupyter/nbformat/issues/170, it should've been fixed a while ago.

yuvipanda:
or https://github.com/jupyter/nbformat/issues/166 maybe.

PR #2190, opened 2021-02-07
Add micro text editor to default image

fperez:
Adds the `micro` command-line text editor as a default, that should be friendlier than vim to newcomers who find themselves in a text editor unexpectedly (such as when making a git commit).

I looked around a lot and [micro](https://github.com/zyedidia/micro) seems like a great way of providing simplicity and ease of use at the command line.  It does what folks new to the CLI are more likely to expect from a text editor (shift-arrow highlight,` `C-x,c,v` for cut/copy/paste, `C-q` for quitting, etc). 

We often find newcomers getting dropped into vim from e.g. a git commit operation, and it can be very jarring. If teaching the CLI in-depth isn't part of a course's goals, this can be an unwelcome distraction/source of confusion.

We're making improvements to JupyterLab to ease this with its own GUI tools, but I think having a super-friendly/familiar editor as a baseline at the CLI for scenarios where teaching vim/emacs isn't viable would be great.

cc @pbstark @Dmacracy in case they have further thoughts.

### DIFF
--- a/deployments/datahub/images/default/Dockerfile
+++ b/deployments/datahub/images/default/Dockerfile
@@ -25,6 +25,7 @@ RUN apt-get -qq update --yes && \
     apt-get -qq install --yes \
             tar \
             vim \
+            micro \
             locales > /dev/null
 
 RUN echo "en_US.UTF-8 UTF-8" > /etc/locale.gen && \


fperez:
ps - if we do this, I'd be inclined to also hard-code the `$EDITOR` env. variable in the system-wide environment config in `/etc` to be `micro` so it would work even for users who haven't configured it. Not getting dropped  into vim/emacs unexpectedly is one of my goals to make hosted unix environments friendlier to newcomers :)

yuvipanda:
Thanks for opening the PR, @fperez. If you wanna set $EDITOR, you should probably put it under `extraEnv` in https://github.com/berkeley-dsep-infra/datahub/blob/13f9509a49687de66d2da6d2b498f4c860720777/deployments/stat159/config/common.yaml#L32. The image is common across all hubs, so don't wanna put it in /etc just yet. We should consider it at the end of the semester though.

fperez:
Thx for the tips, @yuvipanda! Done in #2192, I hope I didn't mess up too much :)

PR #2192, opened 2021-02-08
Update admins and editor defaults

fperez:
This is my first time dealing with the k8 config, so apologies if it's all wrong...

Two things here:

- I'd manually set @Dmacracy as an admin on our hub with the Admin GUI, but I saw here that info is set persistently via this config file. So I imagine the right thing to do is to add him to the yaml file, but please correct me if I'm wrong :) (I can remove that commit if it doesn't make sense).

- As per @yuvipanda's advice in #2190, I've set `micro` as `$EDITOR` by default now. I hope I got the yaml/k8 details right... 

If it works well in general (we'll see how it goes this semester), I think installing `micro` by default and configuring it as `$EDITOR` could be a good idea - it seems like a good CLI tool while being more friendly to users new to the CLI/Unix.  Being dropped straight into vim/emacs out of a git or similar call can be super jarring for someone who's never seen those tools, and I'd like students to have a smoother on-ramp on this (even if later they do end up learning vim/emacs for more complex tasks).

### DIFF
--- a/deployments/datahub/images/default/Dockerfile
+++ b/deployments/datahub/images/default/Dockerfile
@@ -25,6 +25,7 @@ RUN apt-get -qq update --yes && \
     apt-get -qq install --yes \
             tar \
             vim \
+            micro \
             locales > /dev/null
 
 RUN echo "en_US.UTF-8 UTF-8" > /etc/locale.gen && \


yuvipanda:
Did you accidentally close this, @fperez? :)

fperez:
Ugh! No wonder this wasn't working :) I have no idea what happened, I seem to have accidentally nuked it.

Will open a new one now, we actually need this :)

PR #2193, opened 2021-02-08
Add midnight commander (mc) package for better remote file management.

fperez:
mc is a very powerful, yet still command-line-based file manager. Having it
available by default will make it easier for students to do more complex file
operations on the hub than by purely issuing unix commands.


yuvipanda:
Thanks, @fperez.

PR #2202, opened 2021-02-09
Set $EDITOR and add Daniel as persistent admin

fperez:
Remake of #2192 which I somehow nuked inadvertently (and destroyed the branch when I mistakenly thought it had been merged).


yuvipanda:
Thanks, @fperez

PR #2252, opened 2021-03-16
Update numpy to 1.20.1 and add numpydoc.

fperez:
I'm a little worried about changing the main numpy version for others, but for stat159 it turns out we need it, as cryptorandom (which we're working with extensively) has bumped its dependencies to 1.20.x...

I hope it's possible to make this change only for our s159 hub, without any impact/disruption on any of the other hubs...

### DIFF
--- a/deployments/datahub/images/default/Dockerfile
+++ b/deployments/datahub/images/default/Dockerfile
@@ -231,7 +231,7 @@ RUN jlpm cache dir && mkdir -p /tmp/yarncache && \
         @jupyterlab/geojson-extension && \
     rm -rf /tmp/yarncache
 
-RUN pip install --no-cache numpy==1.19.5 cython==0.29.21
+RUN pip install --no-cache numpy==1.20.1 cython==0.29.21
 
 COPY requirements.txt /tmp/requirements.txt
 RUN pip install --no-cache -r /tmp/requirements.txt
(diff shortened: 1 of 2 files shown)


yuvipanda:
@fperez ah, so what timeline do you need this in? Immediately? Or can it wait a day or two?

yuvipanda:
In the meantime, you can add `%pip install -U numpy==1.20.1` to the first cell of your notebook, and that should upgrade numpy when that's run. Won't last across user server restarts though.

fperez:
Thx @yuvipanda!! A day or two is fine, no worries... We'll tell the students to install it locally if need be (that's what I did at first to get through the initial setup.

fperez:
Ping @yuvipanda  - if it's easy to get this one merged this week, we'd greatly appreciate it!

Also - is it possible to update JupyterLab to current (3.0.11) at the same time? Last night I noticed that 3.0.5 (the one we're using) seems to have a specific bug where it ignores Tab indentation in the editor, which is problematic for Makefiles... The problem seems gone in 3.0.11.

If you prefer a separate PR for the JLab change just let me know, happy to do so...

yuvipanda:
@fperez I don't want to bump numpy for everyone, so I'll just make a small custom image. I'll make sure it's done before Friday AM Pacific.

fperez:
I figured as much, sorry for the hassle! Makes total sense, much appreciated...

Any thoughts on the JLab fix? We're working around the bug by having the students use `micro` at the terminal (glad we added that a while back!) but it would be nicer to rely cleanly on the Lab editor if they want a GUI. I haven't tested in detail, but at least on my machine 3.0.11 seems OK with tab indentation, while 3.0.5 in the hub seems to ignore it.

yuvipanda:
@fperez yeah, #2258 should fix lab. I'm working on the image itself as well.

PR #2270, opened 2021-03-29
Add build-essential package (gcc, make, etc)

fperez:
We need at least Make for our core work, I hope this is a reasonable fix.


yuvipanda:
These images inherit from [buildpack-deps](https://hub.docker.com/_/buildpack-deps), which comes with build-essential among other such packages. I can access `make` from datahub on my account - are you running into errors?

fperez:
Yup, just getting `command not found` on `make`. Our workaround for the students right now is to install it via conda (big save there!) but better if it's visible at the user level.

yuvipanda:
Fixed by #2271

PR #2273, opened 2021-03-30
Add JupyterBook to default 159 image

fperez:


yuvipanda:
stat159 has its own image now - at https://github.com/berkeley-dsep-infra/datahub/tree/staging/deployments/stat159/image. I'll add it there, hang on.

fperez:
Ah, thanks, sorry about the confusion! I'll update my bookmark for where to go when making these small tweaks :)

yuvipanda:
I've included your commits in https://github.com/berkeley-dsep-infra/datahub/pull/2274, will merge that when ready.

yuvipanda:
@fperez yw! fwiw, since it has its own image, xeus and friends are available if needed :)

Issue #2242, opened 2021-03-09
Move away from hub db dir for log collection


yuvipanda:
We want to collect long term hub logs, mostly for analytical purposes. They're currently stored in the same disk as the hub db (see https://github.com/berkeley-dsep-infra/datahub/blob/staging/hub/values.yaml#L363), which works out until it doesn't - EECS hub had new user spawns fail due to the disk being full! We monitor hub db disk space in grafana, but that's clearly not enough.

We should instead push logs out to [stackdriver](https://cloud.google.com/products/operations) and then from there to an appropriate sink like google cloud storage.

PR #2310, opened 2021-04-10
Bump datahub hub-dir size to 80Gi

yuvipanda:
The 40Gi disk was full, causing an outage.

Ref https://github.com/berkeley-dsep-infra/datahub/issues/2242


yuvipanda:
I manually edited the PVC, and the directory resized + remounted. This just formalizes it.

PR #2316, opened 2021-04-17
Fix Rprofile.site to point to correct mirror

yuvipanda:
Earlier, interactive use of R somehow wasn't loading
appropriate instructions on setting CRAN from Rprofile.site,
resulting (partially) in errors like this:

https://github.com/berkeley-dsep-infra/datahub/issues/2255#issuecomment-806268236

My informed guess is that the `local{}` was scoping the effects,
so it worked during build but not during interactive use.
Removing that seems to make it work for both.

Inspired by https://github.com/toolforge/paws/pull/62


ryanlovett:
The bit about the user agent string is bonkers! I just read 18.3.3 in the doc you linked to.

Issue #2322, opened 2021-04-20
Grant more CPU to users in eecs hub


felder:
@Quizenger reports that students are experiencing long run times when doing their course work on the eecs hub.  A look at the graphs from yesterday does suggest high cpu utilization.

I'm going to start with CPU guarantee of .25 which I believe should place 32 students on a single node.

felder:
@Quizenger yesterday cluster utilization peaked around 60% as opposed to 100% the day before.

Did your students also see improved performance?

Quizenger:
@felder Yes! The runtime dropped significantly for most students. We still had a few at around 10 mins (down from 20 for most). Do you think we can increase the CPU guarantee further? (We only need this until the end of next week, 4/30)

Thanks so much for your help, appreciate it!

felder:
@Quizenger yeah let's give .33 a shot.  That should take us from ~32 to ~24 users per node.

My only concern here is that peak utilization at a guarantee of 0.25 was 60% which may imply that additional bumps won't do much.  One thing to note is that while we're using a minimum guarantee there is no limit set.  So I'd expect lots of students with CPU bound processes to result in 100% utilization like we saw the other day.

I'll go ahead and do the increase now, but I also think @yuvipanda should maybe look at the utilization tonight and see if we're accomplishing much.  

Additionally please let me know in this ticket tomorrow how things are going.

felder:
Unsurprisingly these changes have resulted in the spinning up of additional nodes.  I modified the autoscaling rules for the gamma node pool to permit a max of 30 nodes instead of 20.

felder:
The gamma node pool is hovering around 18 nodes.  I'm bumping the gamma node pool's max from 30 to 40 nodes.

Just want to be sure we don't run out of nodes if there is a surge of users.

Quizenger:
We're still facing issues with some students facing >15 min wait times.

I assume the gamma node pool is a backup pool of resources in case there are too many users?

yuvipanda:
@Quizenger how has it been today/

Quizenger:
It was better, thank you!

yuvipanda:
https://github.com/berkeley-dsep-infra/datahub/pull/2365 will turn this off.

@Quizenger can you share the profiling info (both the py-spy and top one) here so we can refer back to it next semester?

Quizenger:
[...]

PR #2337, opened 2021-04-26
Add node placeholders

yuvipanda:
We currently use 'pod placeholders'
(https://zero-to-jupyterhub.readthedocs.io/en/stable/administrator/optimization.html)
to try keep 'headroom' of X nodes over what users are currently
using - so when a new user comes in, they can get a node right
away without needing to wait for a spinup. In practice, this has
been problematic - primarily because the unit of placeholder is *pod*,
but the unit of autoscaler is *node*. A common failure mode is

1. A new user comes in, displayes a single placeholder pod. This kicks
   off the autoscaler, and a new node spins up.
2. Another user comes in - but since the new node is emptier than the
   old node, the new user is scheduled in the new node! They have to
   then wait for the image to be pulled, which can take a long time.

So users still see a long wait, even though there's an 'empty'
node.

With the node placeholder approach, we make one pod that tries
to take up *all* the resources on a node. When a user pod comes up,
it can displace this *one* pod - and free up a full node of resources!
The displayed pod will trigger a scale-up, but since not even a single
other user pod can be on the same node as our placeholder pod, the new
user pods will continue to be on the recently vacated node.

So in our case, I've added enough headroom for 3 extra delta
nodes - primarily for use by the EECS hub. This should theoretically
help with a surge of upto 150 users over the course of new node spinup
and ready - about 15 minutes or so.

Let's try it out

Ref #2322

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -234,6 +234,49 @@ jobs:
           command: |
             hubploy deploy eecs hub ${CIRCLE_BRANCH}
 
+  deploy-node-placeholder:
+    docker:
+      - image: buildpack-deps:bionic-scm
+    working_directory: ~/repo
+    steps:
+      - checkout
+
+      - run:
+          name: install google-cloud-sdk
+          command: |
+            curl -sSL https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-265.0.0-linux-x86_64.tar.gz | tar -xzf -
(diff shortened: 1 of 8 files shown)


SanjayB:
Dear Mr Yuvi

Please note i am not the Sanjay Bahl  , you have mentioned in the mail,
please check the mail address again

Thanks

yuvipanda:
I've updated it - sorry, @SanjayB!

PR #2342, opened 2021-04-26
Add rpy2 for the R magic integration

fperez:


fperez:
Ah, I didn't realize R was gone entirely :)

If it's easy to add back basic R support in Jupyter (R kernel suffices, no need for RStudio in full), that would be lovely. But not a huge deal if not. I should have caught this earlier... The course is Python based, but for the last projects we're using some reference R code from published literature. So it would be handy if the students could access both in the same machine (they're likely to have R on their personal boxes so there's a workaround).

I'll keep this in mind for next time (spring 2022) and set it up with R from the start. Lesson learned :)

yuvipanda:
@fperez I can add R back :) What is the timeline for this?

fperez:
We're using R in the current HW and will do so for the final project too (due in about 2 weeks). If it's easy to do it this week, it would be phenomenal. But no worries if not - I didn't think in advance about the need for R availability, so I fully understand if we don't have it... They typically have R locally, so we do have a workaround. 

Thx again!

yuvipanda:
Done in #2360

PR #2441, opened 2021-06-17
Add PH250A Instructor and GSIs to admin privileges

Kristaneumann:
We will be using Datahub for our 2021 Fall class. Thank you!


yuvipanda:
This was merged in a different PR!

PR #2571, opened 2021-08-19
Adding Data 102 Instructors/GSIs as admins and making edits to the Github templates!

balajialg:
**Commit: Making edits to the github templates**
I missed updating few key info and as a result, the new template to update feature requests was not visible in Github. Pushing this commit to verify whether this fixes that issue.


yuvipanda:
Thanks, @balajialg! In the future, it would be nice to have these split across two separate Pull Requests, so they can be reviewed and merged separately more easily.

balajialg:
Makes sense @yuvipanda! I created a commit, made a pull request, and added the second commit. I found that Git clubbed both these commits together into the same pull request when I tried creating a new pull request. Will read documentation to avoid this scenario when I commit next time.

PR #2609, opened 2021-08-24
adding Stat 20 R package configuration

andrewpbray:


yuvipanda:
Thanks, @andrewpbray!

PR #2649, opened 2021-08-30
Adding @cs.berkeley.edu suffix for Data8 admin

balajialg:
Despite this [PR](https://github.com/berkeley-dsep-infra/datahub/commit/7ff6aa9867d8b77cc14b7e13e830f4f78627b838), I found that the user did not get added as an admin in Data 8 hub's staging server. So looked at the previous version of YAML file and figured that for those with @cs.berkeley.edu, their entire email id got added.

### DIFF
--- a/deployments/data8/config/common.yaml
+++ b/deployments/data8/config/common.yaml
@@ -33,7 +33,7 @@ jupyterhub:
           - felder
           - balajialwar
           # Fall 2021 Admins, from Meghan Wang  
-          - daw
+          - daw@cs.berkeley.edu
           - nellepersson
           - meghanwang
           - melissarwong


felder (changes requested):
This doesn't look right to me, namely because no one else has been granted access this way.  I've added @yuvipanda to the reviewer list and hopefully he'll chime in here.  I'm guessing what you need is the first part of their @berkeley.edu email address.

yuvipanda:
Let's ask the user what they see when they log in and open [the control panel](https://data8.datahub.berkeley.edu/hub/home)? On the top right they should see their username.

balajialg:
Closing this PR as I could find that the user had admin access [here](https://data8-staging.datahub.berkeley.edu/hub/admin)! Sorry for the confusion

Issue #2533, opened 2021-08-12
Python package bumps for datahub ahead of Fall 2021


fperez:
**Which package do you want installed?**

These are all installed but with older versions, it would be good to bump up to these current versions:

```
jupyterlab     : 3.1.6
matplotlib     : 3.4.2
numpy          : 1.21.1
pandas         : 1.3.1
plotly         : 5.1.0
scipy          : 1.7.1
sklearn        : 0.24.2
statsmodels    : 0.12.2
xarray         : 0.19.0
```

**Which hub do you want it installed on?  For example r.datahub, datahub, data100.datahub, etc.**

Data 100.

**Which class will use this package?**

Data100 - @fperez and @alvinwan

**Which semester will this package be used for?**

Fall 2021

**Any additional information we should know about?**

JupyterLab >= 3.1.5 is particularly important, so that we get the fix for [this issue](https://github.com/jupyterlab/jupyterlab/issues/10732) that was merged in 3.1.5.

fperez:
For reference, these are the current versions on the 100 hub

```
jupyterlab     : 3.1.4
matplotlib     : 3.3.3
numpy          : 1.19.5
pandas         : 1.2.0
plotly         : 4.14.3
scipy          : 1.6.0
seaborn        : 0.11.1
sklearn        : 0.24.0
statsmodels    : 0.11.1
xarray         : 0.16.2
```

fperez:
In case anyone wants to do a quick check of what's there upon import (rather than checking conda/mamba package lists, to avoid any potential kernel namespace confusions), I use this snippet to quickly see what's in my various environments:

```python
import sys
from importlib import import_module

mods = sorted("""
jupyterlab
matplotlib
numpy
pandas
IPython
scipy
sklearn
seaborn
statsmodels
plotly
xarray
geopandas
""".strip().splitlines())
    
for mod in mods:
    try:
        p = import_module(mod)
        print(f"{mod:<15}: {p.__version__}")
    except ModuleNotFoundError as err:
        print(err, file=sys.stderr)
    except AttributeError:
        print(f"{mod} has no '__version__' attribute", file=sys.stderr)
```

yuvipanda:
@fperez should be all new now!

fperez:
[...]

PR #2665, opened 2021-09-01
Bump JupyterLab again

yuvipanda:
Ref https://github.com/berkeley-dsep-infra/datahub/issues/2533#issuecomment-910469810


fperez:
Awesome, thanks a lot @yuvipanda!!

cc @andrewlenz @abadrinath947 - this should take care of some of the issues we've been seeing, in particular this kind of cell collapse that was proving to be both confusing and frustrating to sort out:

image

PR #2739, opened 2021-09-12
Upgrade to Otter v3.1.3

chrispyles:


yuvipanda:
@chrispyles so the publichealth hub wants to pin ottr to 0.1 for this semester. Will this upgrade affect them? If so, we should pin otter-grader for the publichealth hub...

chrispyles:
@yuvipanda Yes, it will affect that. I'll pin the public health hub to otter v2.2.7.

yuvipanda:
Thanks, @chrispyles. However, I think this will get accidentally overwritten the next time we run `./scripts/infra-packages/sync.bash`. Can you instead explicitly pin it  [here](https://github.com/berkeley-dsep-infra/datahub/blob/6d75ac49d43dbaff2748d38e83e7daf7e5ca3ae3/deployments/publichealth/image/Dockerfile#L59)?

chrispyles:
@yuvipanda done

yuvipanda:
Thanks, @chrispyles

PR #2746, opened 2021-09-13
Post annotation to grafana when node-placeholder scales

yuvipanda:
Helps us see the effects the scaler has on various other
metrics


yuvipanda:
image

Annotations are being posed now!. https://grafana.datahub.berkeley.edu/d/R3O4mbg7z/cluster-information?orgId=1&refresh=30s

PR #2758, opened 2021-09-15
Adding packages requested for Phys-188

### LINKED ISSUE #2757: Request assignment1910 package for phys188/288
**Which package do you want installed?**

Hello. May I request the re-installation of the following phys188/288 code?

https://github.com/berkeley-dsep-infra/datahub/commit/942b4c8c208e380cc288bbdd59b06ab06d964cad#diff-2cb9238ae53e266b72f1de1268e9818a2ee79801c9030f9eaa3eb9b2be785c6f

**Which hub do you want it installed on?  For example r.datahub, datahub, data100.datahub, etc.**
datahub

**Which class will use this package?**
Physics 188/288  
(https://classes.berkeley.edu/content/2021-fall-physics-188-001-lec-001)

**Which semester will this package be used for?**
Fall 2021

Thank you!

balajialg:
Added the packages as per the comment referenced as part of the GitHub issue #2757


deployments/datahub/images/default/Dockerfile
```diff
@@ -242,6 +242,9 @@ ENV NLTK_DATA ${CONDA_DIR}/nltk_data
 COPY connectors/text.bash /usr/local/sbin/connector-text.bash
 RUN /usr/local/sbin/connector-text.bash
 
+COPY connectors/2019-fall-phys-188-288.bash /usr/local/sbin/
```
yuvipanda: The file seems to be missing from the PR?

balajialg: @yuvipanda aah got it. How to attach this file as part of PR? I don't know its location

felder: @balajialg the file was removed.

However, you can view what it used to be by following the URL given in the request:
https://github.com/berkeley-dsep-infra/datahub/commit/942b4c8c208e380cc288bbdd59b06ab06d964cad#diff-2cb9238ae53e266b72f1de1268e9818a2ee79801c9030f9eaa3eb9b2be785c6f

Then clicking on the 3 dots next to the deleted file to get access to former contents:
https://github.com/berkeley-dsep-infra/datahub/blob/1e2f491f49f08bde1a34e41c98cb22f1b86b733b/deployments/datahub/images/default/connectors/2019-fall-phys-188-288.bash

So in this case it'd probably be easiest to just create a new file in the location indicated, paste the contents there into it, chmod 755 it (so it's executable), then git add it, git commit it, then git push it.

balajialg: @felder Thanks a lot for the inputs. Made the corresponding changes and pushed them as part of the same PR

felder (changes requested):
@balajialg note the build log says:

"/bin/sh: 1: /usr/local/sbin/2021-fall-phys-188-288.bash: Permission denied"

This is the likely the result of not setting the file as executable (the chmod 755 step I mentioned in the previous PR).

Looks like another option you can try is git's update-index command:
https://git-scm.com/docs/git-update-index#Documentation/git-update-index.txt---chmod-x
`git update-index --chmod=+x path/to/file`

The executable bit can also be set when the file is added apparently:
`git add --chmod=+x path/to/file`

At this point you would want to do the former.

balajialg:
@felder Aaah, I did chmod 755 for this file. Will try the other options you had suggested to see if they are working. Thanks!

balajialg:
@felder Did as per your suggestions, let's see whether this fixes the issue. Thanks!

felder:
@balajialg the commit shows the file attribute change and build completed, but the build output looks gnarly.  @yuvipanda what do you think?

deployments/datahub/images/default/connectors/2021-fall-phys-188-288.bash
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -euo pipefail
+
+cd ${CONDA_DIR}
+git clone https://github.com/HerculesJack/assignment-1910
+cd assignment-1910
+bash install.sh
```
yuvipanda:
```suggestion
bash install.sh > /dev/null
```

yuvipanda: @balajialg @felder I think this might make the output log a bit better?

I think this is ok to merge, but let's set a 'back out date' when we can remove this. This should be published as a package on pypi.org and installed for the next time - I think we should try to stay away from just running arbitrary scripts.

felder:
@balajialg looks like the warnings are still there, they might be going to standard error.

perhaps we need `bash install.sh > /dev/null 2>&1`

We could also just merge this.  I really think ultimately @yuvipanda is right and we should ask @ybh0822 to package this (or reach out to the responsible party for this) as well as to investigate all the build warnings.

ybh0822:
@felder Thanks for all the help! I sincerely appreciate it. I will request the repo owner to package this module, but I don't think this will be completed anytime soon...
I believe it has not been installed yet? 
I tried:
> sys.path.append('/opt/conda/assignment-1910')
> from cosmofast.planck2018._simall import _simall_f, _simall_j, _simall_fj

and it seems that I can't load the modules.
Thank you so much!

felder:
@ybh0822 no it is not installed yet.  We're still working on it.

ybh0822:
Thank you again!

Issue #2784, opened 2021-09-20
Request more RAM for class ESPM-157 


cboettig:
**Which hub do you want more RAM on?**

 r.datahub

**Which class is this request for?**

https://classes.berkeley.edu/content/2021-fall-espm-157-001-lab-001

**How many students do you expect in this class?**

54

**How much RAM does this class need?**

4 GB

**Why does this class need this much RAM?**

Students reproduce analyses of key global change papers, and need to be able to read in the original scientific data into R.  In particular, we need to be able to load the RA Myers legacy stock assessment database and R isn't so friendly with low-RAM environments.

**Any additional information we should know about?**

balajialg:
@cboettig Thanks for this request! What is the timeline you are looking at? How many weeks/months do you want to use the R hub with the increased RAM allocations.

Such requests have cost implications for us and hence wanted to get the duration details for this request.

cboettig:
Thanks!  We are using reasonably data-intensive exercises in the modules starting today through the rest of the semester (including the student final projects). I totally understand the cost restrictions, and I try pretty hard to keep the assignments within a 4 GB limit and not go over that, but really the difference between what students can do in 4 GB and 1 GB is very substantial.  

Please let me know if that's not feasible and we can try and explore other hosting options; e.g. whether we can deploy your kubernetes system on our own hardware -- we already do this with self-hosted GitHub Actions since the free GitHub Actions for Educational accounts is just way too low to be practical in a large class, even with relatively minimal checks.

balajialg:
Hi, @cboettig Thanks for the detailed rationale for making this request! Are you foreseeing this requirement across subsequent semesters for your courses? 

We don't have a policy around such requests, and we would like to use your request as a model to define our policy. We should be able to get back in a day or two with regards to this request. Would that timeline work for you?

cboettig:
@balajialg Thanks.  Yes, if possible we'd like to keep being able to use r.datahub for the course going forward as well.  We used r.datahub with 4 GB allocations last year during remote instruction (#1849) and it worked really nicely.

balajialg:
@cboettig Thanks for sharing the PR! That's helpful to put things in context. I will keep your requirement for the R hub in mind as we respond to this request.

yuvipanda:
Based on the metric in https://github.com/berkeley-dsep-infra/datahub/issues/2785, let's approve this.

cboettig:
[...]

PR #2841, opened 2021-10-06
Increasing RAM for ESPM student by adding their entry to sops file for issue #2784

balajialg:


yuvipanda:
@balajialg I've just added a comment referencing the issue in the config itself, but otherwise lgtm!

PR #2862, opened 2021-10-08
Installing library tzdata as part of docker image in biology hub for issue #2737

balajialg:


petersudmant (approved):
Assuming it works! Looks good!

yuvipanda:
I think this might have already been installed by default on all our images, as a dependency of other packages?

Issue #2608, opened 2021-08-24
Update PR documentation to include R packages


ryanlovett:
# Template Credits: 2I2C
# Summary

The [docs for submitting a PR](https://docs.datahub.berkeley.edu/en/latest/admins/howto/new-packages.html#submitting-a-pull-request) mention conda, pip, and python, but don't mention R package management. One instructor was uncertain how to proceed. We should probably document the r-packages/ directories and the files it contains.

# User Stories

This is for instructors and GSIs who want to make PRs for adding R packages for their new courses.

balajialg:
Thanks, @ryanlovett, for raising this important issue! If you could support me with relevant information, I could take a stab at getting this out by next month!

FYI, Yuvi created a related [issue ](https://github.com/berkeley-dsep-infra/datahub/issues/2550)sometime back!

PR #2874, opened 2021-10-12
Adding PR documentation to install R packages to resolve issue: #2608

balajialg:


felder (changes requested):
These instructions for installing a package are more complicated than they likely need to be.

Like if the package in question is just missing, install.packages("packagename") is probably the best way to do an initial check to see if the default installation repo contains the package (and version) required.

Also I'm not sure "When testing RStudio with new version of the R package, add the following line to a cell at the beginning of your notebook." really applies here since the work occurs in RStudio typically and not in notebooks.

balajialg:
@felder Removed the unnecessary changes based on your inputs. Let me know whether it suffices.

balajialg:
@felder Any suggestions on the next step for this PR?

felder:
@balajialg This is pretty good.  I don't think we should get much more complicated here such as using specific github commit ids (which is something we do) or some such.

PR #2881, opened 2021-10-14
Update infer and stat20data

andrewpbray:


felder:
I'm going to close this pull request for now.  If it turns out we can update this package without causing issues for ph142, we will submit another PR to do so.

andrewpbray:
Is it ok to reopen this one now since it sounds like PH is not on r.datahub?

felder:
@andrewpbray the fix for infer is still being deployed.

ryanlovett:
@andrewpbray I'm pulling in stat20data update via #2885. infer is now in staging at 1.0.0 via your original specification in stat-20.r and with ph-1.4.2's infer removed.

PR #2885, opened 2021-10-14
Bump stat20data.

ryanlovett:
As in #2881.

### DIFF
--- a/deployments/datahub/images/default/r-packages/stat-20.r
+++ b/deployments/datahub/images/default/r-packages/stat-20.r
@@ -25,7 +25,7 @@ class_libs = c(
 class_libs_install_version(class_name, class_libs)
 
 devtools::install_github("hadley/emo@3f03b11")
-devtools::install_github("stat-20/stat20data@90e67d0")
+devtools::install_github("stat-20/stat20data@48b57a2")
 devtools::install_github("andrewpbray/boxofdata@8afd934")
 
 print("Done installing packages for stat-20")


andrewpbray:
Hmm, when I go on to r-staging, I see the updated `infer` but I don't see the updated `stat20data`. It should have a few spelling changes (the data set `millenials` is fixed to `millennials`). Did I get the version wrong in the req file?

ryanlovett:
Looks like you specified 48b57a2 in https://github.com/berkeley-dsep-infra/datahub/pull/2881/commits/1582ede62e8e0632c492ea806888fc19da54fa4f and that is what got merged in #2885.

This is currently merging to prod.

andrewpbray:
Which is weird, because that's the version that has the update I think?

[https://github.com/stat-20/stat20data/blob/48b57a26bd792d79758647b53f599a01a02b5b22/data/millennials.rda](https://github.com/stat-20/stat20data/blob/48b57a26bd792d79758647b53f599a01a02b5b22/data/millennials.rda)

ryanlovett:
It looks like https://github.com/stat-20/stat20data/commit/48b57a26bd792d79758647b53f599a01a02b5b22 changed two files, man/promote.Rd and man/yawn.Rd. How does one evaluate those from within R?

andrewpbray:
Oh, looks like it all appears correctly now on r-staging. I guess the `infer` PR went through well before the `stat20data` one?

Oh, and now it looks like production is all good to go both with `infer` and `stat20data`! Thanks a ton for your help!

ryanlovett:
Okay, great. Yeah, your additional infer change wasn't necessary since infer 1.0.0 was included in stat-20.r originally. So instead of asking you to reopen and change your PR, I just made this one.

PR #2930, opened 2021-10-26
Allow *all* outbound access to the internet

### LINKED ISSUE #2825: Consider unblocking all outbound ports
The bigger question is to determine what benefit we get from filtering outbound ports, and see if we should consider just removing that as a limitation. Particularly, we allow outbound 443/80 to the wide internet with no restrictions. So, users can in fact tunnel out to wherever they want on the internet via that.

However, there is a lot of value in allowing only explicitly listed ports when accessing internal IPs. For example, there's no reason for user pods to hit promethues server or the configurable-http-proxy API endpoint. So maybe we should try and find a way to allow unlimited outbound internet access but tightly controlled access to things on the internal network?

 I'll try form some thoughts too, but also pinging @ryanlovett and @felder who might have opinions. Removing outbound port restrictions will also fix FTP (#1789), making @cboettig happy.

yuvipanda:
- Disallow all access to internal network except when explicitly
  allowed - so this is all k8s pods, services, nodes and possibly other
  services in the same private network.
- Allow talking to the ingress controller, so hubs can talk to their own
  public URLs - without this, python code running in
  data8.datahub.berkeley.edu can not actually make HTTP requests to
  data8.datahub.berkeley.edu! However, this doesn't work for datahub
  yet, as it uses its own IP with autohttps and not the ingress
  controller.
- Keep data8x restricted on outbound access. This hub is open to the
  wider public, not just Berkeley students.

TODO:

- Unclear if outgoing traffic to statsd is working or not. Needs to
  be investigated.

Thanks to @sgibson91 for
https://github.com/2i2c-org/infrastructure/pull/774,
where parts of this are stolen from

Fixes https://github.com/berkeley-dsep-infra/datahub/issues/2825#event-5429999808

### DIFF
--- a/deployments/data8x/config/common.yaml
+++ b/deployments/data8x/config/common.yaml
@@ -66,6 +66,22 @@ jupyterhub:
           # 2813 Eric Van Dusen
           - bd1e7c64c6f5faddbf6b32af3010cf75 #sean.smorris
   singleuser:
+    networkPolicy:
+      # We only allow outbound HTTP, HTTPS and DNS access
+      enabled: true
+      egress:
+        - ports:
+            - port: 80
+              protocol: TCP
+        - ports:
+            - port: 443
+              protocol: TCP
+        - ports:
+            # statsd ports
+            - port: 9125
+              protocol: TCP
+            - port: 9125
+              protocol: UDP
     initContainers:
       - name: volume-mount-hack
         image: busybox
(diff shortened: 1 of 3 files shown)


yuvipanda:
I can't figure out what's up with statsd tho.

yuvipanda:
Let's deploy this and see if statsd stuff still comes through.

PR #2964, opened 2021-11-04
Pin setuptools before installing JupyterHub

yuvipanda:
The setuptools bug was being triggered by us installing
JupyterHub from git. In
https://github.com/berkeley-dsep-infra/datahub/pull/2963,
we pinned setuptools in the same file as we install JupyterHub -
this is too late as setuptools has already been invoked to install
JupyterHub!

So we install setuptools before we setup infra-packages.


yuvipanda:
Closing in favor of #2965

Issue #2966, opened 2021-11-05
data8 needs more CPU


yuvipanda:
Currently, data8 students are reporting that cells sometimes take really long to run.

<img width="1360" alt="image" src="https://user-images.githubusercontent.com/30430/140467304-dc55ba1a-5f15-4713-8fab-ca80ac8870fd.png">

This is because the CPUs are being maxed out on all data8 dedicated nodes, by users running code at 100% CPU. Since we overprovision here, it doesn't work out well.

yuvipanda:
As there's a project deadline, I've:

- Manually deployed #2967, as we still have a datahub image build happening in CircleCI
- I increased the minimum amount of nodes in the delta nodepool to 13, and scaled up nodes to 20. This should help data8 students who restart to land on a new and existing node. However autoscaler immediately brings it down to 13 anyway, so that wasn't helpful

yuvipanda:
I've also created a new gamma pool with `n1-custom-16-16384` machine type - 16 cores and 16GB of RAM, matching the 0.5 CPU guarantee and the 0.5GB memory guarantee we give users. Previously, they were on n1-highmem-8 nodes, which wasted a *lot* of RAM - only about 16 users could fit on it due to the CPU constraint, leaving a lot of wasted RAM.

yuvipanda:
Let's tone down teh CPU requests on Monday.

yuvipanda:
I've turned this down to a 0.1 guarantee and a 0.5 limit with #2983.

yuvipanda:
I added an alert for 90% CPU utilization in https://grafana.datahub.berkeley.edu/d/R3O4mbg7z/cluster-information?orgId=1&viewPanel=9.

yuvipanda:
<img width="693" alt="image" src="https://user-images.githubusercontent.com/30430/140869981-898e783d-759b-4f55-8493-6a26c5b96501.png">

The CPU requests are filling up the node faster than the memory requests, which I don't want - data8 (and most of our students really) are bottlenecked on memory, not CPU. Since the goal is to make sure students never end up with *no* CPU, I'm instead going to give everyone 0.05 (5%) CPU than 10%

balajialg:
@yuvipanda Closing this issue as the underlying problem seems to be solved by you. Please reopen if that is not the case!

PR #2967, opened 2021-11-05
Give data8 students more CPU

yuvipanda:
Reported on the data8 slack that many users are having trouble running
*any* cells at all. Apparently there is a project deadline today with many simulations,
and a few users have pegged CPU because we don't really have any CPU limits.

The CPU limit and guarantee matching will ensure that no set of users can
take over a full node, and everyone gets the same amount of CPU - only fair
when you are doing homework. I've also created a new nodepool with 16GB of
RAM and 16 cpu cores to better match the CPU / RAM guarantee ratio we have
for our user pods.

Ref https://github.com/berkeley-dsep-infra/datahub/issues/2966

### DIFF
--- a/deployments/data8/config/common.yaml
+++ b/deployments/data8/config/common.yaml
@@ -140,7 +140,7 @@ jupyterhub:
 
   singleuser:
     nodeSelector:
-      hub.jupyter.org/pool-name: delta-pool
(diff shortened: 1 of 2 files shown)


yuvipanda:
Deploying this manually to give data8 users a quick fix, as our previous deploy / build of the datahub image is still ongoing.

Issue #2934, opened 2021-10-27
Install most packages from conda-forge, instead of pypi


yuvipanda:
Currently, we get most of our python packages from pypi.org, installed via pip. A lot of scientific python packages have C extensions, and installing them from pypi has been simple enough thanks to [manylinux wheels](https://www.python.org/dev/peps/pep-0513/). However, there are some packages - particularly in the geo sciences - that are a pain in the ass to install this way still. 

https://github.com/berkeley-dsep-infra/datahub/issues/2824 is one such case. The [cartopy](pypi.org/project/cartopy) project does not ship manylinux wheels, so we need to install its C dependencies - proj, geos, gdal, etc - from apt. This also has knock-on effect for other packages that depend on proj, like [shapely](https://pypi.org/project/Shapely/). It *does* have binary wheels, but because cartopy and shapely must link to the same proj library, it must be built from source too - or you run into problems with https://github.com/berkeley-dsep-infra/datahub/issues/1796.

This becomes even more complicated when we add R to the mix. The [sf](https://packagemanager.rstudio.com/client/#/repos/1/packages/sf) R package also needs proj, and since we're installing it from packagemanager.rstudio.com, it's linked against the version of proj that is available in apt. 

So to recap, the following package managers are involved:

1. apt to get C libraries (proj)
2. pip to get Pyton packages, that link against the C libraries (shapely, cartopy)
3. R packages coming in via packagemanager.rstudio.com, that also link against C libraries specifically coming from apt (sf)

This was a bit of a tenuous situation, but the need to upgrade cartopy for https://github.com/berkeley-dsep-infra/datahub/issues/2824 totally made this unworkable. Cartopy 0.20 needed a newer version of proj than what was available in apt. With https://github.com/berkeley-dsep-infra/datahub/pull/2826, we tried to install a newer version of proj from conda (adding yet another package manager to the mix!), but this required we remove proj installed via apt - as otherwise pip was still trying to link to that, and that doesn't work. And once we removed proj from apt, this broke the R `sf` package, as it required proj from apt!

I think the core of the problem is that both pip and R are dependent on apt for some C libraries, and this can conflict. I propose instead that we:

1. Use conda to get most scientific python packages, especially any that have C dependencies. This completely removes the need for C packages from apt for the most part
2. Use apt to get C packages needed by R packages.

[...]

PR #2977, opened 2021-11-05
Moving dependencies to conda-forge for julia, issue #2934

felder:


felder:
@yuvipanda I didn't move matplotlib from requirements.txt, but I can (or at least I'll check to see if I can).

As I'm sure you could imagine, there were some packages that I could not find in conda-forge.

What I opted to do was to list them in environment.yml, but commented out.  I also did the reverse infra-requirements.txt, that is I commented out the ones I moved and didn't comment out the ones I couldn't.

If this generally looks good, let me know and I'll proceed with a similar methodology for the rest of the hubs.  If there's other stuff related to this task you think I should also do, also please let me know!

yuvipanda:
This is a great start, @felder! 

Since the infra-requirements.txt package is common across all images, let's actually tackle that *last* - it's still useful to bump some packages across the images (like with https://github.com/berkeley-dsep-infra/datahub/pull/2969). Instead, just move what's in requirements.txt to environment.yml?

felder:
@yuvipanda I'm a bit confused here, because in #2934 you commented with "Move infra-requirements.txt to install from conda, by specifying it in environment.yml"

What did you mean by this?

yuvipanda:
@felder ah, so the question is one of order of operations. infra-requirements.txt does need to be moved, but the canonical copy is in https://github.com/berkeley-dsep-infra/datahub/blob/staging/scripts/infra-packages/requirements.txt and https://github.com/berkeley-dsep-infra/datahub/blob/staging/scripts/infra-packages/sync.bash is the script that ensures that they are the same across the images. So if we were to move infra-packages, we'll have to move *that* requirements.txt to an environment.yml file, and write a somewhat more complex script to keep that in sync across the images, as I'm guessing it might not be in multiple environment.yml files. Hence my suggestion here that we move https://github.com/berkeley-dsep-infra/datahub/blob/staging/deployments/julia/image/requirements.txt as that's just specific to the julia image.

felder:
@yuvipanda ok got it.

I'll close this PR and submit another with just requirements.txt moved.

Is this general case for all the rest of the images as well?

yuvipanda:
@felder yeah, I think that's the way to go. We can do infra-requirements.txt last?

felder:
@yuvipanda sure.  I'll set the files aside so that when it is time to approach that, I'll already know which items are in conda forge.

PR #3008, opened 2021-11-16
Moving requirements.txt to environment.yml for biology issue #2934

felder:
Also setting timezone to America/Los_Angeles


felder:
@petersudmant just giving you a heads up that as per issue #2934 I'm folding requirements.txt into environment.yml  I also set the timezone to America/Los_Angeles as we've previously discussed.

felder:
@yuvipanda just checking in with you regarding items conda didn't know about.

https://github.com/berkeley-dsep-infra/datahub/blob/2614e021304c1cb6f0a546ec5fba7e99aef3e6b8/deployments/biology/image/environment.yml#L47

deployments/biology/image/environment.yml
```diff
@@ -35,5 +35,16 @@ dependencies:
 - raxml=8.2.*
 - muscle=3.8.*
 - dendropy=4.4.*
+
+# requirements.txt
```
yuvipanda:
```suggestion
```

Don't think we need to mention this

yuvipanda (approved):
This LGTM!

PR #3011, opened 2021-11-16
Limit biology hub users to 4 CPUs

yuvipanda:
A couple of biology hub users have been saturating the CPU on
the nodes they're on, impacting performance on other hubs in
the beta nodepool (ischool and public health). Bringing them
down to a limit of 4 CPUs should help. If this continues, we can
just move biology to its own nodepool.


yuvipanda:
image

You can see the beta pool nodes maxing out from the biology pods here.

PR #3041, opened 2021-11-19
Moving requirements.txt into environment.yml for data8x issue #2934

felder:


felder:
@yuvipanda can you give this one a clean bill of health before I merge?  It's a bit different than the other hubs in that there is no Dockerfile in the repo, so I just wanted to be sure.  Thanks!

yuvipanda:
Ah cool - it is using [repo2docker](https://repo2docker.readthedocs.io/en/latest/config_files.html) to build the image. This looks ok to me.

yuvipanda (approved):

Issue #3057, opened 2021-11-24
request more Mimi Julia Packages 


ericvd-ucb:
### Package Name

multiple Mimi packages?

### Hub details

- Hub: 
- Julia
- Language: 

we have MIMIFund and its working 
Now we need the others

CSVFiles 
DataFrames 
Distributions 

MimiDICE2010 
MimiDICE2013 

MimiPAGE2009 
MimiRICE2010
Query 
VegaLite 

MimiIWG

ref
https://github.com/davidanthoff/teaching-2019-aere-workshop/blob/master/Project.toml

and 
@lrennels

### Course Name

Data 100 

### Installation Deadline

Nov 30 if possible?

ericvd-ucb:
FWIW this is what we are trying to run
https://github.com/davidanthoff/teaching-2019-aere-workshop

at JupyterStuffPuller Link

https://julia.datahub.berkeley.edu/hub/login?next=%2Fhub%2Fuser-redirect%2Fgit-pull%3Frepo%3Dhttps%253A%252F%252Fgithub.com%252Fdavidanthoff%252Fteaching-2019-aere-workshop%26urlpath%3Dlab%252Ftree%252Fteaching-2019-aere-workshop%252F

yuvipanda:
The packages that start with MIMI should be added here: https://github.com/berkeley-dsep-infra/datahub/blob/b8e68c02c5ba6108afa2e689e01a714f360a2ffb/deployments/julia/image/install-julia-packages.jl#L22

Rest look like standard packages and should be added here: https://github.com/berkeley-dsep-infra/datahub/blob/b8e68c02c5ba6108afa2e689e01a714f360a2ffb/deployments/julia/image/install-julia-packages.jl#L5.

@ericvd-ucb while this is happening, you should be able to do something like
```julia
using Pkg;
Pkg.add.([<list-of-packages>])
```

to be able to temporarily test.

ericvd-ucb:
@yuvipanda  thanks so much - yes indeed they can all be installed via Pkg.add - but its was kind of laggy to thanks for the installs

felder:
@ericvd-ucb @yuvipanda I've submitted a PR to staging, but there's an issue.  Triangle will not co-exist with MimiFUND, MimiIWG, and Optim.

Triangle also hasn't really been updated much since Julia 1.2 which I suspect has something to do with it.

ericvd-ucb:
Lets drop that one then

[...]

PR #3059, opened 2021-11-30
Adding Julia Dependencies

felder:
Issue #3057

I could not get Triangle to co-exist with MimiFUND, MimiIWG, and Optim.  I suspect this has to do with Triangle's age.

If you attempt to Pkg.add.("Triangle"), Julia will downgrade MimiFUND, MimiIWG, and Optim.  After which MimiFUND and MimiIWG will fail the Pkg.precompile() step.

### DIFF
--- a/deployments/julia/image/environment.yml
+++ b/deployments/julia/image/environment.yml
@@ -5,3 +5,5 @@ dependencies:
 
 # from requirements.txt
 - matplotlib==3.4.3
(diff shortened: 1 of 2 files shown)


felder:
@yuvipanda @ericvd-ucb on slack indicated it's ok if Triangle is dropped.  Given that, I'll send this PR through.

PR #3066, opened 2021-12-03
Removing requirements.txt for datahub

felder:
Issue #2934


deployments/datahub/images/default/environment.yml
```diff
@@ -29,4 +26,248 @@ dependencies:
...
+
+# data100, geog88
+# https://github.com/berkeley-dsep-infra/datahub/issues/2838
+  - pysal==2.5.*
+  
+# cogsci88;
+  - ggplot==0.11.5
+
+# pymc3 needs this
+  - mkl-service==2.4.0
```
yuvipanda: https://github.com/berkeley-dsep-infra/datahub/pull/2544 explicitly installed this from conda-forge, and I think we should keep it there. mkl (https://en.wikipedia.org/wiki/Math_Kernel_Library) is a non-python math library, and the description on https://pypi.org/project/mkl-service/ says to install from conda. Can you try importing the package from the built image? If that works maybe we're clear...

felder: @yuvipanda sadly it does not work.

So I'm moving mkl-service back to conda and doing more trial and error and dependency juggling.

yuvipanda:
I think this could also be something with docker layers duplicated. I have two suggestions on investigating this:

1. Use https://github.com/wagoodman/dive to look at the built docker image, and see if there's a particular layer that's super wasted
2. Run `du` in the conda notebook environment to figure out which packages are taking up most of the space, and compare it with same in the current existing image.

PR #3069, opened 2021-12-07
staging branch to use gofer_service with otter

sean-morris:


felder:
@sean-morris Would you be opposed to resubmitting this PR using a new branch so that only your changes are included?

PR #3070, opened 2021-12-07
Changed gofer_nb url to gofer_service with otter grader

sean-morris:
The gofer_nb url in common.yaml now points to the updated
gofer_servce that is using otter grader


felder:
I'm going to close this because modifying common.yaml will cause this change to affect both staging and prod once the change is merged to prod at some point in the future.

Instead I made a similar change to staging.yaml which will only impact staging and is what I believe you want to have happen.

See PR #3071

Issue #2945, opened 2021-10-29
Plan for Spring 2022


yuvipanda:
Issue to track 

**Things we want to do *before* Spring 2022**

- [x] Support RStudio 2021.09  (They switched to calendar versioning, https://www.rstudio.com/products/rstudio/release-notes/)
- [ ] Have alerts that trigger whenever there is an outage
- [ ] Move the user images out of this repository, to make self-serve image requests easier. See https://github.com/2i2c-org/justiceinnovationlab-image for an example workflow - primary advantage is ability to test the image without needing staging. 
- [ ] <s>Switch to Python 3.10 as default, from 3.9 (should be done during the semester switch)</s>
- [x] Make data8 use retrolab #2422 
- [x] Codify cleanup tasks (https://github.com/berkeley-dsep-infra/datahub/issues/2946)

**Things we want to do *during* Spring 2022**

I## Big Ticket items

- [x] - Ensure that Data 8 adopts Retro Lab? (@yuvipanda)
- [x] - Get stat159 to pilot RTC hub? (@yuvipanda)
- [x] - Upgrade R & RStudio (@yuvipanda @ryanlovett)
- [ ] - Initiate cloud cost-sharing conversations with EECS and (possibly) Biology? (@balajialg)
- [x] - Finalize a decision on resourcing forward - 2I2C pilot and/or resource sharing with RTL (@ericvd-ucb, @balajialg?)
- [x]    - Scoping the JD irrespective of the path taken wrt resourcing
- [ ] - Figure out a storage solution (@felder)
- [ ] - Keep leadership at Berkeley and RTL aware of the changes happening in accessibility support in Jupyter @balajialg 

## Smaller ticket items
- [ ] - Invest more efforts into improving alerting system
- [ ] - Invest more efforts into improving grafana dashboards
- [x] - Communicate our work to our users in the monthly newsletter format
- [x] - Support faculty at Public Policy, Civil, and Sociology adopt Datahub?
- [x] - Evangelize the nbgitpuller extension for Chrome and Mozilla as part of our conversation with the faculty

yuvipanda:
Another potential inclusion here is to move away from SIS and use only CanvasAuthenticator, but I think the other stuff here is higher priority.

balajialg:
[...]

PR #3090, opened 2021-12-20
Bump R versions in biology hub

yuvipanda:
- Bumps to R 4.1
- Bumps to latest RStudio
- Bumps to latest R Kernel

Ref https://github.com/berkeley-dsep-infra/datahub/issues/2945


yuvipanda:
I've manually built and pushed this image.

PR #3098, opened 2021-12-31
add ay250 sp2022; relax and fix other requirements

profjsb:
This PR adds requirements for AY250 (Spring 2022) and relaxes the requirements of other classes, many of which were pinned to specific versions. `okpy` and `allensdk` are the worst offenders of pinned versions which cause other packages to fail to install.


yuvipanda:
@profjsb what do you think of me making an astro.datahub.berkeley.edu with its own smaller image?

profjsb:
That would be fine with me. I'll need some guidance as to how to do that (and add students to that image)

yuvipanda:
@profjsb I'm back to work tomorrow, will set it upt!

yuvipanda:
Closing in favor of discussing this in #3105

yuvipanda:
The main datahub image is a bit of a mess...

PR #3112, opened 2022-01-06
adding astro packages

profjsb:
Use Python 3.9.* to play nice with pytorch. Adding other data(science) tooling packages to the environment.yml. Use conda-forge for some of the complex packaging but pip for most.

### DIFF
--- a/deployments/astro/config/common.yaml
+++ b/deployments/astro/config/common.yaml
@@ -42,3 +42,32 @@ jupyterhub:
     memory:
       guarantee: 512M
       limit: 1G
+
+  # read by z2jh.get_config() in hub/values.yaml
+  # formerly jupyterhub.hub.extraConfigMap
+  custom:
+    profiles:
+      # Shared dataset volumes for astr 250
+      # See https://github.com/berkeley-dsep-infra/datahub/issues/827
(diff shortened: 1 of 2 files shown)


yuvipanda:
@profjsb you generally prefer pip over conda? Works ok with me, but just curious as I'm trying to reduce mixing of packages wherever possible.

yuvipanda:
I've pushed a commit moving some of the config around to account for the fact that we're using a different authentication system here (chained via datahub.berkeley.edu rather than directly to bcourses). So everyone on the hub gets more RAM, rather than just students enrolled in a course.

profjsb:
> @profjsb you generally prefer pip over conda? Works ok with me, but just curious as I'm trying to reduce mixing of packages wherever possible.

Few reasons: conda-forge packaging updates often lag pip in time (sometimes by many days). More important: the conflict resolution with conda-only packaging seems to take much longer than with pip, at least for this combination of packages we need.

profjsb:
Thanks @yuvipanda -- I defer to your decisions around config changes. Please let me know when we're pushed to prod.

yuvipanda:
@profjsb great! deployed now :)

Issue #3139, opened 2022-01-18
Getting added as admin


lenjb:
### Your Name

Leila Njee Bugha

### Affiliated Course Name

ENVECON 118

### Role

GSI

### Name of the Hub

Datahub/ENVECON-118-SP22

### End Date

06/01/2022

felder:
This has been merged to prod.

PR #3147, opened 2022-01-18
[ENVECON 118] Adding GSI as an admin user for Datahub as per request from the issue #3139

balajialg:


felder (approved):
looks good to me

PR #3149, opened 2022-01-18
Remove postgres container for data100

yuvipanda:
Users reporting errors pulling that image - the
particular tag listed seems to be missing from gcr.io


felder:
Related to issue #3225

PR #3150, opened 2022-01-18
Make stat20data library available in shared path.

ryanlovett:


yuvipanda:
@ryanlovett ah, hmm - given there are 600 students, I think a separate hub might actually be cleaner than this. What do you think?

ryanlovett:
@yuvipanda I think using a shared directory is best for how the instructor wants to update the library, but let me know your concerns regarding the class size. Maybe we can chat about this during the strategy meeting.

ryanlovett:
This was abandoned in favor of a new stat20 hub.

PR #3151, opened 2022-01-19
Move most package management to conda/mamba.

fperez:
This PR may not work as-is b/c I directly edited the requirements.txt file and
minimized it to loading only a few things not available on conda-forge.  This
is different from what the team was doing before, so it may cause issues I'm
not aware of.

@yuvipanda happy to iterate on this to make sure it works well with what you want to do with the other hubs, sorry if my changes cause any issues there. What I'm trying to do is minimize the mismatch between book usage/build, local usage and hub usage, so I can teach the students how to do basically the same thing across all three places.

### DIFF
--- a/deployments/stat159/image/infra-requirements.txt
+++ b/deployments/stat159/image/infra-requirements.txt
@@ -16,7 +16,7 @@ jupyterlab-link-share==0.2.4
 nbconvert==6.4.0
 retrolab==0.3.16
 nbgitpuller==1.0.2
-jupyter-resource-usage==0.6.0
+jupyter-resource-usage==0.6.1
 # Matches version in images/hub/Dockerfile
 jupyterhub==1.5.0
 appmode==0.8.0
--- a/deployments/stat159/image/infra-requirements.txt
+++ b/deployments/stat159/image/infra-requirements.txt
@@ -1,34 +1,9 @@
-# WARNING: Original source at scripts/infra-packages/requirements.txt
-# PLEASE DO NOT EDIT ELSEWHERE
-# After editing scripts/infra-packages/requirements.txt, please run
-# scripts/infra-packages/sync.bash.
+# A few packages not available on conda-forge. The rest of the environment is 
+# provided in environment.yml
 
-# This file pins versions of notebook related python packages we want
-# across all hubs. This makes sure we don't need to upgrade them
-# everwhere one by one.
-
-# FIXME: Freeze this to get exact versions of all dependencies
-notebook==6.4.5
-# FIXME: Conflicts with jupyter-server-proxy somehow
-jupyter-client<7.0
-jupyterlab==3.2.4
-jupyterlab-link-share==0.2.1
-nbconvert==6.1.0
-retrolab==0.3.15
-nbgitpuller==1.0.2
-jupyter-resource-usage==0.6.0
-# Matches version in images/hub/Dockerfile
-jupyterhub==1.5.0
-appmode==0.8.0
-ipywidgets==7.6.3
-otter-grader==3.1.4
-jupyter-tree-download==1.0.1
-git-credential-helpers==0.2
-# Enough people like this, let's load it in.
-jupyter-contrib-nbextensions==0.5.1
-jupyter_nbextensions_configurator==0.4.1
 # Measure popularity of different packages in our hubs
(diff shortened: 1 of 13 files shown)


deployments/stat159/image/infra-requirements.txt
```diff
@@ -1,34 +1,9 @@
-# WARNING: Original source at scripts/infra-packages/requirements.txt
-# PLEASE DO NOT EDIT ELSEWHERE
-# After editing scripts/infra-packages/requirements.txt, please run
-# scripts/infra-packages/sync.bash.
+# A few packages not available on conda-forge. The rest of the environment is
```
yuvipanda: Ah, so infra-requirements.txt is managed globally with https://github.com/berkeley-dsep-infra/datahub/tree/staging/scripts/infra-packages - so we can keep it in sync across all the images. It would be great if we can leave it unmodified, as the next run of the script will mangle all these changes.

yuvipanda (changes requested):

yuvipanda:
The image build has also failed due to conflicting requirements in conda. I've mostly just specified packages we directly want without trying to specify dependencies too (unlike a lock file) to make that process easier.

deployments/stat159/image/infra-requirements.txt
fperez: Yup, I saw that and was sure it would require some iteration, thx...

I guess the question is - given you're moving more to conda packages, how would you prefer to handle this idea of defining the complete env in a conda `.yml` file for each course, while having this global one?

I'm happy to iterate on ideas, but I didn't want to mess with the global file to fit the packages I was after for this class...

fperez:
Mmh, I seee... It worked locally, but I guess there's an issue on linux with `gdal`. That library is a PITA. I'll remove it for now - I use in general the same approach you indicate, to only list the actual targets we want and not all deps, this env file was highly pruned by hand to remove all lower-level deps. But I figured I'd leave GDAL in as I know it to be cantankerous. Maybe should have left it out, I guess :)

I'll try that now, thx!

fperez:
Argh, build failed again:

```
Encountered problems while solving.
Problem: nothing provides gdal 1.11.4 needed by fiona-1.6.3-np110py27_0
```

Fiona seems to be a dep of pyproj. I'm going to start nuking all this for now so we can get going with the class tomorrow, and we can figure out a gdal solution later on...

deployments/stat159/image/infra-requirements.txt
yuvipanda: @fperez yeah, we haven't figured out how to move this through quite yet.

For now, do you think you can just leave out packages that exist in the global file from the stat159 environment.yml? I think the global file should be uptodate on versions.

fperez:
And now it seems the RISE conda-forge package is borked. Argh. Nuking that too, trying again.

fperez:
Mmh, now I give up - this error makes no sense to me:

```
Encountered problems while solving.
Problem: nothing provides jupyterlab >=3.0.0rc10,==3.* needed by jupyterlab-system-monitor-0.8.0-pyhd8ed1ab_0
```

Especially as the previous built had gone through the conda part. @yuvipanda any ideas here?

fperez:
OK, for now it seems with these changes the rest passes. Yuvi, I'll commit next the updates to the master infra-reqs file so you can evaluate them, but I'm happy to take a different approach if you prefer.  They'll go in the next commit without anything else.

yuvipanda:
Wow, thank you so much @fperez! I think biology hub currently fails to build due to timeouts, but I'll do that manually. Will deploy this once build succeeds.

fperez:
Thanks @yuvipanda! I just noticed I'd missed one more package, just pushed one tiny update so we're fully synced to start the term :)

fperez:
Thanks for the help - with this and #3154 I think we should be good to go. Sorry for not gettting this earlier, but recent times were a bit mad courtesy of the world bug :)

fperez:
Just added `micro` as per your comment in #3154. I think this does it, whew...

yuvipanda:
Thanks a lot, @fperez! I might have to build some of these images manually, will do that shortly.

fperez:
Hugely appreciated @yuvipanda!!

PR #3154, opened 2022-01-19
Add an apt.txt file with a few CLI tools that will be handy for Stat 159

fperez:
As suggested by @yuvipanda - I think these should do the trick, thx!

### DIFF
--- a/deployments/stat159/image/apt.txt
+++ b/deployments/stat159/image/apt.txt
@@ -1,0 +1,18 @@
+# Some linux packages for basic terminal work, particularly
+# oriented at newer users
+
+# A couple of CLI editors that are easier than vim
+micro
+nano
+
+# Basic Emacs configuration for general development.
+emacs-nox python-mode emacs-goodies-el
+
+# powerful terminal-based file manager, better than the one in JLab
+mc
+
+# Regular build tools for compiling common stuff
+build-essential gfortran
+
+# Dependencies for nbconvert
+texlive-xetex texlive-fonts-recommended texlive-plain-generic
--- a/deployments/stat159/image/apt.txt
+++ b/deployments/stat159/image/apt.txt
@@ -6,13 +6,18 @@ micro
 nano
 
 # Basic Emacs configuration for general development.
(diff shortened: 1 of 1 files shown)


fperez:
Huh?? I'm puzzled - for [the jupytearth image](https://github.com/pangeo-data/jupyter-earth/blob/c88a5bfab8bda1adf8373d12ab58e169aa56b75f/hub.jupytearth.org-image/Dockerfile#L31) we install `micro` just fine, but this build is [saying it can't find it](https://app.circleci.com/pipelines/github/berkeley-dsep-infra/datahub/989/workflows/4d5ae9d6-9a6d-4212-9f3f-8a35f320f4cd/jobs/8120?invite=true#step-108-974):

```
E: Unable to locate package micro
```

I'm not sure why, as it [supposedly works](https://github.com/zyedidia/micro/issues/583#issuecomment-687035485). Puzzled...

yuvipanda:
@fperez oh that is strange. I think the packages need to be one per line. Can you try that?

Repo2Docker is also still using Ubuntu 18.04, maybe micro was added after that?

fperez:
Ah, that's possible - I just put the packages one per line. But you're probably right about the 18.04 issue, as [per this comment](https://github.com/zyedidia/micro/issues/583#issuecomment-605624736).

fperez:
[Here](https://snapcraft.io/install/micro/ubuntu) they suggest installing it as a snap:

```
sudo snap install micro --classic
```

Should that go into a postBuild file then?

yuvipanda:
@fperez `postBuild` doesn't have admin, so we can't use that. 

For now, we can install it with conda from conda-forge: https://anaconda.org/conda-forge/micro. How does that sound?

fperez:
And for now I'll just tell them to install it themselves :) 

```
curl https://getmic.ro | bash
```

fperez:
Ah, cool, yes, even better!

fperez:
I didn't realize conda-forge had it, thx!!

fperez:
I'll update #3151 then to pull it in too.

PR #3155, opened 2022-01-19
Add clipboard support

fperez:
`micro` complains its clipboard support won't work without xclip/xsel installed.


yuvipanda:
The clipboard itself still won't work, because it's running in a remote VM - xsel / xclip only work on local machines. It might shut off the warning though. If you still want it, am happy to merge!

fperez:
Ah, understood - might as well to disable the warning, thx!

Issue #2813, opened 2021-09-30
Add Morris and Van Dusen to Data 8x hub


ericvd-ucb:
### Your Name

Eric Van Dusen 

### Affiliated Course Name

Data 8x

### Role

admins

### Name of the Hub

- 

### List of Usernames

- 

> 

> 

### End Date

-

balajialg:
@ericvd-ucb @sean-morris Can you follow the below steps and share your EdX Id? Need that for providing admin access!

EdX IDs look like the random strings eg: c741d11d7f0bed9930e04276115b1a37. 

You can find yours by:

1. Enrolling in the class
2. Launching into the lab
3. Finding the user id from the control panel

sean-morris:
Here you go: 
bd1e7c64c6f5faddbf6b32af3010cf75

Thanks

balajialg:
@sean-morris Thank you! Once @ericvd-ucb  shares his id will create a PR for admin access!

balajialg:
@sean-morris FYI, We have provided admin access!

sean-morris:
great thanks. @balajialg

balajialg:
@ericvd-ucb Closing this issue considering the urgency! Please feel free to reopen based on your need.

PR #3175, opened 2022-01-21
Please add me to admin list

sean-morris:
I put this on staging; I am not sure if this is correct for such a request.

I am trying to add myself to hub admins so I can help research problems with notebooks


balajialg:
@sean-morris user id provided by you got added to the admin list as per this issue - https://github.com/berkeley-dsep-infra/datahub/issues/2813#issuecomment-932611034. Is the id starting with 2 another id you want to add?

sean-morris:
Ahhh... This is very confusing to me. So I gave you :
bd1e7c64c6f5faddbf6b32af3010cf75

What account is that!

Let me double-check myself. This is killing me.

balajialg:
Merging this PR based on @sean-morris's observation below,
_the one that was in the admin list is associated with 8.1x. the one I just submitted is associated with 8.3x_

PR #3190, opened 2022-01-25
Add Ruby and jed editor to base image.

fperez:
Not urgent at all, just some extra conveniences as I keep digging into making the system a "home away from home" for all dev work.

### DIFF
--- a/deployments/stat159/image/apt.txt
+++ b/deployments/stat159/image/apt.txt
@@ -1,6 +1,9 @@
 # Some linux packages for basic terminal work, particularly
 # oriented at newer users
 
+# Basic unix tools
+man
+
 # Download tools
 curl
 wget
--- a/deployments/stat159/image/apt.txt
+++ b/deployments/stat159/image/apt.txt
@@ -8,6 +8,7 @@ wget
 # A couple of CLI editors that are easier than vim
 # micro  # currently not working on 18.04
 nano
+jed
 
 # Basic clipboard support
 xclip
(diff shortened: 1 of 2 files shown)


yuvipanda:
Looks like another failure to resolve the environment.yml file: https://github.com/berkeley-dsep-infra/datahub/blob/staging/deployments/stat159/image/environment.yml. Only suggestion I have is to try simplify the environment.yml file, and not specify deeper dependencies (like nbformat, for example) that are already brought in by other packages. This should reduce the amount of possible conflicts.

fperez:
This is so weird, the error we got before was:

```
Encountered problems while solving.
Problem: package sphinx-4.4.0-pyh6c4a22f_1 requires importlib-metadata >=4.4, but none of the providers can be installed
```

which makes no sense to me - sphinx is a very vanilla package, I have no idea why it would have problems resolving. Do you think we could be seeing bugs in conda or issues with conda-forge? This is odd...

fperez:
In any case, I added `man` as you suggested on Slack and tried to trim the env file a bit further.  But it's not much, at this point most of the packages listed are actually the ones we want, so there's not a ton of internal dependencies left in there. I'd already tried to prune that as much as possible.

yuvipanda:
@fperez build has succeeded now! I'll get this deployed. I will admit that I don't know enough about how to debug conda solve failures...

fperez:
Thx @yuvipanda! It's kind of concerning to see these weird conda solve errors, b/c they don't make much sense, but they also seem rather non-deterministic. That environment is absolutely OK to build/install, sphinx is a completely normal package without any weirdness to it, so it's very, very bizarre.

Oh well, a fight for another day! thx again!!

Issue #3198, opened 2022-01-25
Add Admin Access to Data 100 Hub


Dominicliu129:
### Your Name

Dominic Liu

### Affiliated Course Name

Data 100

### Role

TA

### Name of the Hub

- Data 100

### List of Usernames

- yanlisa
- hug@cs.berkeley.edu

### End Date

- 05/20/2022

balajialg:
@Dominicliu129 Instead of cs.berkeley.edu id, Can I use j.hug@berkeley.edu id for creating admin credentials?

Dominicliu129:
@balajialg Hi, I think Prof. Hug uses h___@berkeley.edu for datahub, can you add this instead? Thank you!

balajialg:
@Dominicliu129 Let's add the cs.berkeley.edu account for now (considering the PR is merged in prod). If the prof has issues with accessing the admin instance, I will create another PR with your suggested email id.

Dominicliu129:
@balajialg Hi! I don't think yanlisa has been added to the datahub admin? Can you check again?

balajialg:
@Dominicliu129 Yes, already added. Can you ask the user to restart the server and confirm the same?

PR #3199, opened 2022-01-25
Spring 2022: Adding Data 100 instructors as admins as per issue #3198

balajialg:


felder (changes requested):
Please resolve the indicated merge conflicts.

Also, doesn't #3198 mention hug@cs.berkeley.edu?  I'm not sure if that'll work since authentication eventually ends up being by calnet id.

@ryanlovett @yuvipanda Do you know?  Will the username portion of @cs.berkeley.edu addresses work for adding admin access as requested in #3198 ?

balajialg:
@felder Resolved the merge conflict! Requested the TA for instructor's berkeley.edu email id. Prior conversations related to this issue can be seen [here](https://github.com/berkeley-dsep-infra/datahub/pull/2649) which highlights that this shouldn't be a problem. Let me know if you think otherwise

felder:
@balajialg I was looking for that!  I vaguely recalled this coming up before.

I did look to see if I could find them in the user list (as a non admin) but I could not.  Note that in the conversation you linked, it could be that particular user happened to use the same name for both their calnet ID and the cs.berkeley.edu address.

I guess if they find they don't have admin rights, we can ask them to check what the system says their name is as yuvi suggested.

balajialg:
@felder Makes sense!

PR #3205, opened 2022-01-26
Add more command-line tools

fperez:
The key one here is really the man pages support, please LMK if you prefer that I separate b50ec74 from the other commit(s). The others were just me adding various tools as I think more long-term, but then I ran into this for every single command:

```
jupyter-fernando-2eperez[~]> man ls
No manual entry for ls
See 'man 7 undocumented' for help when manual pages are not available.
jupyter-fernando-2eperez[~]> man bash
No manual entry for bash
See 'man 7 undocumented' for help when manual pages are not available.
```

and I realized we need the actual _content_ of man pages, not just the `man` command :)

No big rush though, I can do without for now.

And sorry I didn't catch all these issues earlier!! Please don't think any of this is urgent - I deeply appreciate the team's work and I'm totally fine now with the setup we have. This is now polish for the long run, not a critical need.

### DIFF
--- a/deployments/stat159/image/apt.txt
+++ b/deployments/stat159/image/apt.txt
@@ -3,6 +3,10 @@
 
 # Basic unix tools
 man
+man-db
+manpages-posix
+manpages-dev
+manpages-posix-dev
 
 # Download tools
 curl
--- a/deployments/stat159/image/apt.txt
+++ b/deployments/stat159/image/apt.txt
@@ -12,20 +12,18 @@ manpages-posix-dev
 curl
 wget
 
+# Core text editors on a *nix box: vim and emacs
+vim
+emacs-nox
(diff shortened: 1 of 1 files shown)


yuvipanda:
@fperez I really appreciate the PRs you are making :)

I opened https://github.com/jupyterhub/repo2docker/pull/909 in repo2docker to allow us to use newer base images, maybe I should pick that up again...

fperez:
Certainly @yuvipanda - I'm delighted with how the system is shaping up, and I'm trying to really "make it my home" so we can reliably use it for all daily work. Your team's recent improvements are phenomenal and are getting us there, I just try to do my small part :)

We're getting to a pretty awesome setup, so this is now polish on top. Exciting!

PR #3239, opened 2022-02-04
add viz packages; fix tensorflow

profjsb:


profjsb:
@yuvipanda any updates on this? I need to get this into prod for a class assignment.

yuvipanda:
@profjsb I can also give you merge rights here, so you can self-merge changes to the astro image. How does that sound?

profjsb:
Thanks. Can you restrict merge rights to just astro-related charges? Too much power is scary!

Issue #3260, opened 2022-02-15
JLab reverted to 3.2.8 -why?


fperez:
### Bug description

In #3246 we updated JupyterLab to 3.2.9 to correct a small but important bug. I see today that we're back to JupyterLab 3.2.8, is there a reason for this reversion? 

Tomorrow's lecture was precisely going to use the console and was the reason for the 3.2.9 update, so I'm curious as to why it got reverted.

Thanks for any clarification!

### Environment & setup

Stat 159 Hub.

### How to reproduce

Log in, open JupyterLab - 3.2.8 is the current version.

felder:
Logging into stat159 I can confirm via the terminal:

```
(notebook) jovyan@jupyter-felder:~$ conda list | grep jupyterlab
jupyterlab                3.2.8                    pypi_0    pypi
jupyterlab-drawio         0.9.0              pyhd8ed1ab_0    conda-forge
jupyterlab-geojson        3.2.0              pyhd8ed1ab_0    conda-forge
jupyterlab-git            0.34.2             pyhd8ed1ab_0    conda-forge
jupyterlab-link-share     0.2.4              pyhd8ed1ab_0    conda-forge
jupyterlab-variableinspector 3.0.9              pyhd8ed1ab_0    conda-forge
jupyterlab_pygments       0.1.2              pyh9f0ad1d_0    conda-forge
jupyterlab_server         2.10.1             pyhd8ed1ab_0    conda-forge
jupyterlab_widgets        1.0.2              pyhd8ed1ab_0    conda-forge
(notebook) jovyan@jupyter-felder:~$ 
```

Looks like while jupyterlab was specified in environment.yml, the version being used is being installed via pip.  Looks like the culprit is likely:

deployments/stat159/image/infra-requirements.txt

Which has jupyterlab 3.2.8 specified.

fperez:
Ah, thanks!! I forgot about that file, it means my PR was incomplete, basically...

I'll make a new one right away fixing the infra file too, thx.

fperez:
Done in #3261, thx @felder!

felder:
Looks reasonable in staging:

```
(notebook) jovyan@jupyter-felder:~$ conda list | grep jupyterlab
jupyterlab                3.2.9              pyhd8ed1ab_0    conda-forge
jupyterlab-drawio         0.9.0              pyhd8ed1ab_0    conda-forge
[...]

PR #3261, opened 2022-02-15
Update JupyterLab 3.2.8 -> .9 to fix console bug

fperez:
This completes #3246, which applied the change only to the environment.yml file.


fperez:
Closes #3260.

fperez:
We'll redo it for only 159.

PR #3287, opened 2022-03-03
Disable RTC on stat159 hub

yuvipanda:
https://uctech.slack.com/archives/CC3NKQA3B/p1646276416672839
@fperez encountered serious data corruption bugs, that
we suspect is caused by enabling collaborative editing.
Let's turn it off and see if it goes away?

### DIFF
--- a/deployments/stat159/config/common.yaml
+++ b/deployments/stat159/config/common.yaml
@@ -39,10 +39,6 @@ jupyterhub:
       # Tell code where to display GUIs
       # The VNC /desktop link must be opened already for this to work
       DISPLAY: ":1.0"
-    # Enable RTC and jupyterlab-link-share
-    cmd:
-    - jupyterhub-singleuser
-    - --LabApp.collaborative=true
     extraFiles:
       lab-config:
         data:


fperez:
Just filed [a JLab issue](https://github.com/jupyterlab/jupyterlab/issues/12154) so we can track it further there with the Lab team.

fperez:
Thanks a lot @yuvipanda for the fast response!

balajialg:
Thanks, @yuvipanda for quickly looking into this! Sorry, @fperez that you had to face these issues when the classes are going on. 
One question to think about is - If RTC seems to be the root cause for all these issues, then will that be the end of the road for testing RTC this semester? I know that we will report this issue in the respective channel but wondering whether the benefits of re-enabling RTC (if the issues are fixed) is worth the risks associated with data loss.

fperez:
Probably, @balajialg, though we should evaluate carefully. For stat159 specifically there is _huge_ benefit in having RTC, and I was very much looking to use it a lot. But after this week's problems I'm very worried.  Let's see what the JLab team's response and feedback are - if a clear cause is identified and corrected, we could do some light trials - I have a separate research hub where I work with my research students and colleagues (that @yuvipanda has access to) and we could do some testing there.

So I'm not quite sure yet - I'm torn between really wanting it for valid pedagogical reasons, and being very worried about potentially losing student work.  If there's any lingering doubt of that happening I'll have to be conservative, so let's just keep a close eye on this and see how things evolve.

balajialg:
@fperez Makes sense! I agree that losing students work poses to be a huge risk so need to be conservative with regards to the final decision. Good to know that there is a plan B (testing it out in the research hub) if choosing to be conservative is our choice! Fingers crossed with regards to JLab team's response.

PR #3299, opened 2022-03-09
Add pytest-notebook

fperez:
Inspired by [@yuvipanda's recent addition of pytest-notebook support](https://github.com/jupyterhub/repo2docker-action/pull/83), I figured I should teach it in 159 too!


fperez:
Awesome, thx  @yuvipanda!!

PR #3333, opened 2022-04-02
Remove duplicate files

yuvipanda:
They were both .md and .rst files


yuvipanda:
As an FYI, now there are rendered documentation previews! If you expand the checks and find the 'docs/readthedocs.org' check, it'll take you to https://readthedocs.org/projects/uc-berkeley-jupyterhubs/builds/16540892/, which has a 'view built docs' link on the right (once the build completes). Clicking that, you can see https://uc-berkeley-jupyterhubs--3333.org.readthedocs.build/en/3333/ - that's how docs.datahub.berkeley.edu will look once this is merged! Helpful to test things out. @balajialg

balajialg:
@yuvipanda Fantastic, previewing doc would be helpful because I see a delta in navigation options between the locally rendered ones and the ones built by readthedocs. The option to move to the next page was not available in certain pages with the local rendering (while appearing in this version)! 

On other note, I did git rm on all the rst files that were not required and pushed the commit, yet these files still got added to this commit which makes me wonder whether I did the git rm accurately.

yuvipanda:
@balajialg do you think you can write up a little info on how to build the docs locally as well as how to use the preview here?

balajialg:
@yuvipanda Yes. added to my to-do list

Issue #3338, opened 2022-04-10
Outdated TexLive + Missing Package


geneh0:
### Bug description

Students are unable to knit to PDF and get the error below. 

I don't think it's actually the outdated TexLive, but somehow `amsmath.sty` is missing now. 

```
A new version of TeX Live has been released. If you need to install or update any LaTeX packages, you have to upgrade TeX Live.

tlmgr: Local TeX Live (2021) is older than remote repository (2022).
Cross release updates are only supported with
  update-tlmgr-latest(.sh/.exe) --update
See https://tug.org/texlive/upgrade.html for details.
! LaTeX Error: File `amsmath.sty' not found.

! Emergency stop.
<read *> 

Error: LaTeX failed to compile lab08.tex. See https://yihui.org/tinytex/r/#debugging for debugging tips. See lab08.log for more info.
In addition: Warning messages:
1: In system("timedatectl", intern = TRUE) :
  running command 'timedatectl' had status 1
2: In system2("tlmgr", args, ...) :
  running command ''tlmgr' search --file --global '/amsmath.sty'' had status 1
Execution halted
```

I ran `kpsewhich amsmath.sty` and there is no output. 

### Environment & setup

- Hub: Public Health
- Language: R

### How to reproduce

Try knitting to PDF via the one-click interface.

geneh0:
@cdbeon Just so you're aware too

ryanlovett:
In the stat20 hub, which like publichealth also uses geospatial as the base Docker image, we install/update LaTeX packages with `tlmgr` [in the Dockerfile](https://github.com/berkeley-dsep-infra/datahub/blob/ee455e05a86182fb33474887cf43385e8581755f/deployments/stat20/image/Dockerfile#L77).

felder:
@geneh0 can you please provide some steps I can use to reproduce this issue?

Alternatively I can make the change described by @ryanlovett to upgrade ansmath and let you know when it has been deployed to the staging environment so you can test.

felder:
@ryanlovett actually I think stat20 is also likely to fail.

Apparently texlive updates every year in early April:
https://www.tug.org/texlive/

I think the right solution is to instruct tlmgr to use the previous repo.
https://www.tug.org/historic/

Trying the following in staging:
`tlmgr repository add https://ftp.math.utah.edu/pub/tex/historic/systems/texlive/2021/tlnet-final`
[...]

PR #3340, opened 2022-04-11
Telling tlmgr to use 2021 repo for texlive and installing packages to…

felder:
… speed up knitting.

Issue #3338


felder:
Opened issue #3342 for ottr which is blocking this PR from succeeding.

PR #3370, opened 2022-05-03
Added grader dns to data8xv2-prod

sean-morris:
The grader service is set to use grader-prod.data8x.berkeley.edu


balajialg:
@sean-morris It would be a good idea to tag @felder whenever you create a PR related to Data 8x!

sean-morris:
> @sean-morris It would be a good idea to tag @felder whenever you create a PR related to Data 8x!

@balajialg yes ok good deal -- I have been slacking after I submit. I will start tagging; it seems easier. I am working this out locally before we merge this.

Issue #3376, opened 2022-05-07
Generate storage reports of users across all hubs!


balajialg:
# Summary

@felder reported data about the amount of storage that gets used across different hubs. He also generated reports for the varied number of users who used a lot of storage at a per hub level. We found that more than 10 users were storing more than 100 gigs of data which easily accumulates to greater than 1 TB of data. I have documented a [few scenarios](https://docs.google.com/document/d/1KsMmRq4rf40kEazQYwEgFpaz__N7-yauG9nvN-3C3lw/edit?usp=sharing) where users stored more than 100 gigs of data which includes storing genomic data for research purposes, storing climate data for assignments, etc. Interestingly, many such users did not even log into their hub instance during the past six months. I am not completely sure about the cost implications for us. However, it would be valuable to generate monthly reports of users across all hubs who use more than 100 GB of data (or any metric that is meaningful to us) so that we can further do due diligence.

Obviously, the long term automation scope would be to send alerts to users and the admins about users with storage greater than a specific metric in a particular hub who did not log into the hub for the past 6 months (as per our policy defined) 

# User Stories

As a `hub administrator`, I want to `identify inactive users who are storing a large amount of data in their home directories` on a per month basis

# Important information
Storage across different hubs!

3.2 TiB [##########] /eecs
    2.2 TiB [#######   ] /genomics
    1.3 TiB [###       ] /stat159
    1.2 TiB [###       ] /data8
  435.6 GiB [#         ] /data102
  344.4 GiB [#         ] /dlab
   92.4 GiB [          ] /prob140
   84.8 GiB [          ] /cs194
   80.8 GiB [          ] /stat20
   57.6 GiB [          ] /astro
   33.8 GiB [          ] /julia
   20.7 GiB [          ] /buds-2020
    7.6 GiB [          ] /workshop
    6.7 GiB [          ] /stat89a
  408.0 KiB [          ] /xfsconfig
    4.0 KiB [          ]  -
    4.0 KiB [          ]  wat
@   0.0   B [          ]  highschool
@   0.0   B [          ]  biology
e   0.0   B [          ] /test

# Tasks to complete

- [x] Create monthly storage reports based on the criteria defined in a format that is easy to produce

yuvipanda:
https://rawgit.com/zevv/duc/master/doc/duc.1.html seems helpful for stuff like this

balajialg:
**Copy-pasting this for continuity here**

[...]

PR #3377, opened 2022-05-07
Update archival policy for Datahub users

balajialg:
I would like to add an additional policy proposal that involves not archiving storage for users who exceed a certain storage threshold while being inactive beyond 6 months. You can read more about the rationale as part of this issue #3376

This would be an additional change to our existing archival policy document. I also updated our new archived data 
storage timeline to be around 6 months post which the data gets deleted (based on our sprint planning discussion on 5/5).

Open to all your comments to clarify the language and make a robust policy that helps communicate to our users clearly while also bringing cloud costs down.

### DIFF
--- a/docs/policy/storage-retention.rst
+++ b/docs/policy/storage-retention.rst
@@ -10,8 +10,8 @@ Policy
 Criteria
 --------
 
-No non-hidden files in the user's home directory have been modified in
-the last 12 months.
+No non-hidden files in the user's home directory that have been modified in the last 6 months.
+The archival process will not happen for users who store more than 100 GB worth of data in any of our hubs. 
 
 Archival
 --------
@@ -24,22 +24,26 @@ Archival
 
 Rationale
 =========
-
-Today (6 Feb 2020), we have 18,623 home directories in datahub. Most of
-these users used datahub in previous semesters, have not logged in
+Most of our users in per course hubs who used datahub in previous semesters, have not logged in
 for a long time, and will probably never log in again. This costs us
 a lot of money in disk space - we will have to forever expand disk space.
 
-By cleaning it up after 12 months of non-usage, we will not affect any
+By cleaning it up after 6 months of non-usage, we will not affect any
 current users - just folks who haven't logged in for a long time. Archiving
 the contents would make sure people still have access to their old work,
 without leaving the burden of maintaining it forever on us.
 
+Yesterday (5 May 2022) - We found many scenarios where users stored more than 100 GB worth of data (approximated to be around ~2 TB) 
+which includes datasets in the home directories of their hubs. Most of these users did not log into their instances of the hub for the past 6 - 12 months. 
+It seems reasonable to not archive their storage data considering that their a) storage amount is large in comparision with their peers
+ and b) they have not been actively using this storage during the past 6 months.
+By following this practice, we will ensure that we are not spending unnecessarily on cloud storage costs. 
+Once identified, Users should get informed to back up their data in the next 30 days.
+Post which either the data gets deleted from their home directories or is not archived for future retrieval.
+
 Why Google Drive?
 =================
 
-For UC Berkeley users, Google Drive offers `Unlimited Free Space
-<https://bconnected.berkeley.edu/collaboration-services/google/bdrive>`_.
 We can also perform access control easily with Google Drive.
 
 Alternatives
@@ -53,4 +57,3 @@ Alternatives
    link to the users. We will need to probably clean this up every 18 months
    or so for cost reasons. This is the viable alternative, if we decide to
    not use Google Drive
-
--- a/docs/policy/storage-retention.rst
+++ b/docs/policy/storage-retention.rst
@@ -11,7 +11,6 @@ Criteria
 --------
 
 No non-hidden files in the user's home directory that have been modified in the last 6 months.
-The archival process will not happen for users who store more than 100 GB worth of data in any of our hubs. 
 
 Archival
 --------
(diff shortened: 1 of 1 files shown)


yuvipanda:
So this means there are two processes:

1. Run a script, find users who have home dirs >100GB, send them an email. What should this email say? Will we delete their homes, or just not archive them until it goes under 100?
2. Run our achival script, as usual.

Is that right?

balajialg:
Hey @yuvipanda, My suggestion for the email would be to stress the point that we will delete their data if they don't back up in the 30 days.  @felder @ryanlovett any suggestions here?

If we don't archive any users' home directories, by when do their files get deleted? If it stays for a longer duration of time (greater than 6 months), it makes sense to delete the files after giving users a generous window to download their files. If it is a short duration, not archiving makes sense to me. Open to all your thoughts on the best way forward, 

This is the email template I used in my recent communication with users who stored files with sizes greater than 100 GB. I specifically wanted to understand why they chose to store greater than 100 GB so that I can include some pointers to staff during the onboarding email during Fall 2022!

_I am reaching out to you with a couple of requests related to the amount of storage you have in your Datahub instance. We found that you have stored more than 100 GB worth of data in your home directory which is far higher than the storage required by most of our users.

1. Can you tell us if there are specific reasons for storing such a large amount of data?
2. Can you also back up your files locally in the next 30 days? We plan to clean your home directories by the end of a 30-day timeline as higher storage is correlated with higher costs at our end. Note that you may not be able to retrieve those files post the mentioned timeline._

yuvipanda:
Ok, so I think this is a different policy than the user storage *archival* policy. It sounds like the proposed policy would be:

1. If your home directory gets over 100GB, we will email you. 
2. You have 30 days to bring it under 100GB, **or we will delete all of it**

Does this sound right? I'm happy with it as a policy (with tweaks for amount of notice, and how to get exceptions), but I think it's a 'large home directories' policy distinct from our archival policy and we should write it as such.

Can you make this into a new file and title it as a 'home directory size policy' or similar?

THANK YOU SO MUCH FOR WORKING ON THIS!

felder:
@balajialg I'd like to use # of days instead of # of months as a general case.  So instead of 6 months, lets specify 180 days.  Instead of 12 months lets specify 365 days.

Also there may be race conditions between discovering directories that are too large and archival.  Probably not a big deal, really, but it's possible we may archive larger directories if the directories happen to grow between the last size check and the archival process.

@yuvipanda One thing I've noticed, which is probably not surprising, is that all the standard tier persistent disks take a really long time to do size calculations on.  Like it may take over a day to run those calculations on all of the existing storage.  Not a huge deal since I can probably deprioritize the process (which will make it take longer), but something to consider.

balajialg:
@yuvipanda I liked the idea of developing a separate policy for storage limits and @felder will keep in mind your articulation around defining limits in terms of the number of days instead of months. 

One thing I realized during my discussion with @felder is the perverse incentives created  **in the long run,** by defining the storage limit as 100 gigs for every user. We may be allowing scope creep with our communication to the instructors that anything lesser than 100 GB would be a reasonable limit (yes, we can choose not to communicate this policy which is a reasonable pathway).  

What are our goals with regards to storage? Allow our users to run computationally intensive workflow appropriate to their needs with minimal cost spending for unused storage. Obviously, the primary action point is to remove the outliers who are **storing enormous amounts of data for their coursework-related/research-related stuff without accessing their storage regularly**. Just by this action, we will be able to reduce almost 2 TB of unused storage, which can approximately save us up to $2000 - $2500 per year. (Obviously, we can debate whether these are worthy cost savings for the project if the cost of man-hours invested to optimize is greater than this) So, so this in theory seems like a reasonable short-term action with better returns.

However, thinking about storage from a long-term perspective, Setting 100 gigs may not be a reasonable limit considering that the storage needs may vary depending on the nature of the course/dataset used/type of use cases used, etc.. which makes it hard to generalize across hubs. Hypothetically, For genomics, a storage limit of 100 GB could be a reasonable limit but for a political science course, 5 GB could be the reasonable limit. One of the other important problems it creates is the edge cases Eg: If the user has a home directory storage of 97 GB - will our policy apply in such scenarios? Do we still consider their storage as over the limit and include them as part of the storage limit policy? Should we reduce the storage limit further? This seems like a [Nash equilibrium](https://en.wikipedia.org/wiki/Nash_equilibrium) problem. 

One suggestion is to define a storage policy that is slightly dynamic for both per course hubs (Data 8, Data 100, potentially I school hub, etc..) and generic hubs (Datahub, R hub, Julia Hub, etc..). Open to debate whether the arbitrary number defined below makes sense. 

1. At a per-course hub level which may include computationally complex use cases, we can identify the median size of all the user home directories and evaluate whether the user's storage is below 3x the median size of all home directories pertaining to their hub. (This could also be learned based on interaction with faculty). 
3. At a generic hub level, which is expected to consist of foundational use cases, setting the limits that users' home directories should be lesser than 2x the median size of all home directories specific to the hub. 

Let me know what you all think about the above policy suggestions? Obviously, we can debate whether we need to optimize at this level for this policy? Are the efforts involved to define the policy worth the cost savings?

yuvipanda:
@balajialg What do you think of just starting with a 100GB limit that we enforce, and see how that goes? I understand that what we want to express is not 'each student can use upto 100GB', but I worry about added complexity of defining limits for one user based on usage pattern of other users. This becomes messy in big hubs like datahub as you mention, as there isn't a clear way to delineate who should be 'comparable users'. I think simplifying policies so they are the same across hubs is a good way to start, and we can make things more complex if needed.

balajialg:
@yuvipanda Sounds reasonable. I had a conversation with @felder about the archival process based on a few data points he was able to retrieve about the home directories. I will incorporate that as part of the storage policy PR. It might open up a new discussion on the archival process. For now, I will remove the entries related to the storage policy in this policy doc and make a new commit.

yuvipanda (approved):
Looks good to me!

PR #3480, opened 2022-06-24
data8xv2: Added loadRoles to gofer_nb service

sean-morris:
Correct loadRoles config added to common.yaml so
gofer_nb can read users


sean-morris:
I meant to PR from staging -- blew that

PR #3481, opened 2022-06-24
data8xv2: Added loadRoles to gofer_nb service

sean-morris:
- Correct loadRoles config added to common.yaml so
gofer_nb can read users


felder:
@sean-morris Yuvi requested I review this, but I imagine it should be ok because this the new stack which is not in production yet.  However, before I approve it, I just want to be sure.

sean-morris:
Yep not in production yet. The Auuhentication in 2.0 is requiring some
configuration

felder (commented):
Looks ok to me.

Issue #3487, opened 2022-06-27
Request retrieval of my Datahub files


lbiral:
### Email Id

lbiral@berkeley.edu

### Link to Datahub Folder

[- Eg: gs://ucb-datahub-archived-homedirs/spring-2021/datahub.berkeley.edu/xxx.tar.gz](https://datahub.berkeley.edu/hub/login?next=%2Fhub%2Fuser%2Flbiral%2Ftree)

felder:
@lbiral 

I need the link that is in the file that led you here.

lbiral:
> @lbiral
> 
> I need the link that is in the file that led you here.

Let me know if this one works: https://datahub.berkeley.edu/hub/login?next=%2Fhub%2F

felder:
@lbiral no.  I need the link from the document in your home directory.  It's of a similar format as the one in the issue description.

lbiral:
I'm sorry, but I am so confused. Which document in my home directory? I can't access my home directory because my DataHub is shut down. If it's not this one (https://datahub.berkeley.edu/user/lbiral/tree), then I have no idea.

felder:
@lbiral Login via https://datahub.berkeley.edu

Requesting file retrieval only makes sense if you have some indication your files have been archived.  The way we inform you is by placing a file in your home directory which contains instructions to come here and tell us which archive to retrieve.

lbiral:
I never saw a file placed in my home directory. I tried logging in the other day and I got a message saying "Canvas doesn't have an account for user: 1558987" 

I have a lot of important personal and professional files on there so I would really love it if I could get them back.

felder:
We're going to update the issue template to talk about this, but yeah the idea behind archive retrieval requests is that we periodically archive files and inform people that this has been done via a file we place in their home directory.  That file contains instructions to open an issue here and provide the link in that file that tells me which archive to retrieve.

If someone opts to open an issue for file retrieval and was not directed to do so by the above process, then it causes a fair amount of confusion if they attempt to construct their own links.  That sends me on a wild goose chase of sorts and results in confusion.

So I think the first issue to look at here, is why you're unable to login.  It's either a technical issue which we need to address or it's because you're no longer a student at UC Berkeley.  If it's the latter, can you please tell me what courses you took that used datahub and what semesters you took them?  If it's the former, that's the issue we need to address.

lbiral:
[...]

PR #3490, opened 2022-06-29
Get additional information as part of data archival request template

balajialg:


.github/ISSUE_TEMPLATE/data_archival_request.yml
```diff
@@ -8,13 +7,31 @@ body:
...
+      required: true
+  - type: input
+    attributes:
+      label: Term/Year
+      description: Identify the term of the semester (Fall/Spring/Summer) and the year of your course enrollment
+    validations:
+      required: true     
+  - type: input
+    attributes:
+      label: Login Detail
```
yuvipanda, line 31: can i ask why we are asking this? The identifier provided in the text file should be all that is needed for this, right? It also encodes the name of the hub that is archived, so it *should* say which hub it is.

yuvipanda (changes requested):

.github/ISSUE_TEMPLATE/data_archival_request.yml
balajialg: @yuvipanda - My understanding is that these changes will help us have relevant data to respond to the archival request when we encounter scenarios like this - https://github.com/berkeley-dsep-infra/datahub/issues/3487. @felder Please feel free to share more details if required.

felder: @yuvipanda The problem is a number of people open file retrieval requests without being directed to do so by the text file.  Thus far most, but not all, of those cases involve folks who have graduated.  There's also often a mix of archived and unarchived files from different hubs/filesystems for these people.  As @balajialg mentioned issue #3487 is a very good example of this issue.  This person had graduated, could no longer login, and had a mix of archived directories as well as directories I had to manually archive and send to them.

Thus I'm asking @balajialg to include language in the template that can steer folks into giving me the correct info.  

@balajialg I'd like to meet with you to discuss.  Like I don't think we should ask if people have graduated specifically.  Instead, I think we should ask something along the lines of whether or not they can still login and direct them to do so if they can.  If they cannot, I'll need information about what hub(s) they used.

yuvipanda: thank you for pointing out to that issue @felder and @balajialg! Makes sense that is the wrong form to use for this. I'd like to keep this form simple - maybe we can have a different one for people who no longer have access? I'm not sure how this user found this form, but it feels like the problem was that they were completely in the wrong place.

felder:
@yuvipanda when you hit "new issue" in the issue tracker, it's an option in the list.  Seems like that's the most likely way they find it.

I'm definitely open to the idea of having separate issue templates for this.  Even better if we can hide the template associated with the expected use case of being directed here via the text file.

.github/ISSUE_TEMPLATE/data_archival_request.yml
felder: @yuvipanda when you hit "new issue" in the issue tracker, it's an option in the list. Seems like that's the most likely way they find it.

I'm definitely open to the idea of having separate issue templates for this. Even better if we can hide the template associated with the expected use case of being directed here via the text file.

balajialg: @felder Yes, find some slot in my bcal next week to discuss this more. Let's think it through from an user perspective and find an optimal solution :)

PR #3529, opened 2022-07-29
Pinned code-server to 4.5.1

sean-morris:
The vscode distribution on the data8xv2 hub is installing a
two year old version of vscode. This updates vscode to May 2022 - 1.68

### DIFF
--- a/deployments/data8xv2/image/environment.yml
+++ b/deployments/data8xv2/image/environment.yml
@@ -1,6 +1,3 @@
-channels:
-- conda-forge
-
 dependencies:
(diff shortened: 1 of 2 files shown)


yuvipanda:
I think this is because nodejs is pinned to 14 upstream in repo2docker https://github.com/jupyterhub/repo2docker/blob/main/repo2docker/buildpacks/conda/environment.yml#L5. My suggestion now is to run the appropriate conda command in the `postBuild` file, and I'll look at unpinning nodejs upstream next week.

Issue #3605, opened 2022-08-20
Refresh packages and add a few more for D100 - Fall 22


fperez:
### Package Name

Here is a list of the packages I'd prefer to have updated for the hub, with their current versions:

```
IPython                  : 8.4.0
PIL                      : 9.2.0
bokeh                    : 2.4.3
bs4                      : 4.11.1
cartopy                  : 0.20.3
cython                   : 0.29.32
dask                     : 2022.8.1
geopandas                : 0.11.1
h5py                     : 3.7.0
intake                   : 0.6.5
ipycanvas                : 0.12.1
ipyleaflet               : 0.17.0
ipympl                   : 0.9.1
ipywidgets               : 7.7.2
jedi                     : 0.18.1
jupyter_book             : 0.13.0
jupyterlab               : 3.4.5
jupyterlab_favorites     : 3.0.0
jupyterlab_git           : 0.39.0
jupyterlab_lsp           : 3.10.1
jupyterlab_myst          : 0.1.6
jupyterlab_widgets       : 1.1.1
lxml                     : 4.9.1
matplotlib               : 3.5.3
matplotlib_inline        : 0.1.6
myst_nb                  : 0.13.2
myst_parser              : 0.15.2
networkx                 : 2.8.5
numba                    : 0.55.2
numpy                    : 1.22.4
pandas                   : 1.4.3
pep8                     : 1.7.1
plotly                   : 5.10.0
pooch                    : v1.6.0
pyarrow                  : 9.0.0
pyflakes                 : 2.4.0
pylint                   : 2.14.5
pylsp                    : 1.5.0
pytest                   : 7.1.2
pytest_cov               : 3.0.0
pytest_notebook          : 0.6.1
requests                 : 2.28.1
scipy                    : 1.9.0
seaborn                  : 0.11.2
skimage                  : 0.19.3
sklearn                  : 1.1.2
sphinx                   : 4.5.0
sphinx_jupyterbook_latex : 0.4.6
statsmodels              : 0.13.2
sympy                    : 1.10.1
xarray                   : 2022.6.0
yaml                     : 6.0
zarr                     : 2.12.0
```

Most of these are on the hub, but not all - so this request is for a mix of a version refresh for the major packages relevant to D100, and the installation of a handful more that will be needed this semester.  If I run the same check on the current hub, I get:

```
IPython                  : 8.4.0
PIL                      : 8.3.2
bokeh                    : 2.3.3
bs4                      : 4.9.3
[...]

PR #3609, opened 2022-08-22
Switch data100 to using the stat159 image

yuvipanda:
@fperez is teaching it this semester, so let's give him the familiar
environment!

Ref https://github.com/berkeley-dsep-infra/datahub/issues/3605


fperez:
Huge thanks @yuvipanda !

PR #3632, opened 2022-08-25
Update *all* packages in stat159

yuvipanda:
This uses latest version of all available packages, except
for python.


fperez:
Quick q - I thought the conda version syntax required one `=`  instead of two `==`; does it work either way?

yuvipanda:
@fperez ya it works either way I think! If the build fails I'll update my script.

I love this script!

fperez:
Good to know! And please share that script, I want to see it :) I'd started to write the same (couldn't help myself). Much better you did - I can focus on tomorrow's class!

Issue #3679, opened 2022-09-02
Xelatex Installation for EECS Hub


ee120-staff:
### Bug description

[Using this issue template because the "Package addition" template doesn't seem to fit this situation]

Can we get Xelatex installed on the EECS hub? I'm trying to export an otter-grader notebook to PDF and I'm getting
```
OSError: xelatex not found on PATH, if you have not installed xelatex you may need to do so. Find further instructions at https://nbconvert.readthedocs.io/en/latest/install.html#installing-tex.
```
Thank you!

### Environment & setup

- Hub:  EECS Hub
- Language: Python

### How to reproduce

1. Go to https://eecs.datahub.berkeley.edu/hub/user-redirect/git-pull?repo=https%3A%2F%2Fgithub.com%2FEE120-Course-Staff%2Flab-0-introduction&urlpath=tree%2Flab-0-introduction%2Flab0.ipynb&branch=main
2. Run the imports and the bottom `grader.export()` cell.
3. See the error

ee120-staff:
If possible, can this be done before we release our first lab on 09/09?

felder:
xelatex installed!

PR #3680, opened 2022-09-02
Adding package texlive-xetex

felder:
Issue #3679


felder:
@yuvipanda any ideas here on what we can do to not play whack a mole with the package dependencies?  Namely why isn't apt just pulling these in?

PR #3704, opened 2022-09-07
Increase RAM guarantee for data100

yuvipanda:
More kernel deaths are being reported, and looking at memory usage, the beta nodes where users end up seem to have a fairly high memory use %. So users could be getting their kernels killed more frequently than in other hubs. This changes the overcommit ratio quite a bit, so kernels are killed less often


yuvipanda:
image

here is the graph of memory usage on a per-node basis

PR #3752, opened 2022-09-14
Set options(HTTPUserAgent) to contain RStudio Server.

ryanlovett:
In testing, this causes binary packages to be installed in R sessions outside of RStudio. Hopefully this takes effect during image builds and speeds them up.


ryanlovett:
This didn't speed up the image build, but it still improves the experience of installing packages outside of RStudio's console.

PR #3825, opened 2022-10-10
Merge #3823, #3824, #3826, #3827, #3828 to prod

yuvipanda:


hub/templates/nfs-pvc.yaml
```diff
@@ -29,4 +29,37 @@ spec:
   resources:
     requests:
       storage: 1Mi
+---
```
ryanlovett, line 32: If `useV3` is true, won't there be two pv/pvc? (should the v4 block be in an `else` structure?

ryanlovett (commented):
See comment -- should the v4 block be part of an `else`?

hub/templates/nfs-pvc.yaml
ryanlovett: Also, when I try to start a pod on staging I get `pod didn't trigger scale-up: 6 persistentvolumeclaim "home-nfs-v3-v3" not found`.

yuvipanda: @ryanlovett yeah there will be two. We can't modify PVCs or PVs after creation, so to use multiple options we need multiple ones. And we can't delete PVCs when they are still in use, so if we delete the old ones we'll have to kick out *all* users from the cluster

Issue #3850, opened 2022-10-17
Request more memory for Data 100


andrea2031:
**Which hub do you want more RAM on?**

datahub

**Which class is this request for?**

Data 100/200 Fall 2022 (https://classes.berkeley.edu/content/2022-fall-data-c100-001-lec-001)

**How many students do you expect in this class?**

1200

**How much RAM does this class need?**

4GB

**Why does this class need this much RAM?**

We have a high memory usage homework.

**By when (MM/DD) do you want this request to be fulfilled?**

10/17 for admins to be able to start developing, 10/18-11/1 for students if it works

**Any additional information we should know about?**

balajialg:
@andrea2031 Thanks for the info. Till when (MM/DD) do you want the increased compute?

rohanjha123:
Until 11/2 if possible please!

balajialg:
@rohanjha123 Sounds good!

balajialg:
@andrea2031 You can access the increased RAM of 4GB when you restart your server and launch a new notebook. Please reopen if you face any issues. Thanks for your patience.

balajialg:
@rohanjha123 Can we reduce the compute for this hub (from 4 GB to old limits) based on your previous comment? @andrea2031

andrea2031:
Our late due date is tonight 11:59:59pm, it can be closed after midnight today. Thank you again for all your help! Edit: see below

balajialg:
That sounds good @andrea2031. Tagging @felder as he should be aware of this,

andrea2031:
Hi, so sorry we have a couple special extension so they'll need the memory till tonight. Could you push the date of reducing computation resource back a few days (11/4)? Thanks so much!

felder:
@andrea2031 is that midnight on the 4th? If so that implies we should roll back the additional memory on the 7th since I don't think this warrants night/weekend work.

andrea2031:
Rolling back to the daytime of 7th sounds good! Thank you so much again!

PR #3853, opened 2022-10-17
Increasing memory limit and guarantee for Data 100 hub!

balajialg:
For issue #3850


balajialg:
@felder User wrongly reported as Datahub in the issue but it should be Data 100 hub.

felder (changes requested):
Why does this commit want to do 1000+ changes?  Do you perhaps need to "git fetch upstream staging" "git pull upstream staging" prior to branching?

Also it looks like you did not remove the old values for the limits and guarantees.

PR #3889, opened 2022-10-28
give myself admin access

shaneknapp:
after committing this i believe i will have the power!  XD


felder (changes requested):
You probably also want to add yourself to the new hub template here:
datahub/deployments/template/{{cookiecutter.hub_name}}/config/common.yaml

Other than that, looks good to me.

yuvipanda:
@shaneknapp yep, this LGTM! You should self-merge this, then merge staging -> prod to deploy to prod hubs.

PR #3930, opened 2022-11-08
Updating a11y Hub image and pointing it directly in hubploy

balajialg:


deployments/a11y/image/environment.yml
```diff
@@ -0,0 +1,91 @@
...
+- pip
+- pip:
+  - -r infra-requirements.txt
+  - jupyter-desktop-server
+    # For push authentication to GitHub
+#  - gh-scoped-creds==4.1
+  # Upgrade separate from what everyone else uses for now
+  # https://github.com/berkeley-dsep-infra/datahub/issues/3693
+  - otter-grader==4.0.1
+  - ipython-sql==0.4.1
```
ryanlovett: You probably don't need ipyparallel, pyarrow, ipython-sql, pandas, numpy, etc. unless you want to evaluate their UIs for accessibility.

shaneknapp:
lgtm, but i would like confirmation that i didn't miss anything.  ;)

deployments/a11y/image/apt.txt
```diff
@@ -0,0 +1,91 @@
...
+# oriented at users new to Unix/cmd line environments.
+
+# Basic unix tools
+man
+man-db
+manpages-posix
+manpages-dev
+manpages-posix-dev
+
+# Download tools
```
ryanlovett, line 11: A lot of the UNIX cli tools and X11 apps probably aren't needed for Lab accessibility testing, e.g. emacs, nano, jed, mc, zsh, nnn, most of the packages that start with "x", firefox, gedit. Maybe curl/wget and vim are useful for just Doing Stuff, but most everything else won't be useful for testing.

ryanlovett (commented):
Just as an aside, you can test the image on your own device with hubploy before waiting to see what issues shake out in CI. Once you've got hubploy, docker, etc. available locally, it is often faster than the change/deploy/CI cycle.

deployments/a11y/image/environment.yml
balajialg: Thanks, @ryanlovett I will keep pandas and numpy as they are commonly used packages, and can test a few sample notebooks using them. Rest of the packages - I will remove them in the upcoming commit.

deployments/a11y/image/apt.txt
balajialg: @ryanlovett Sounds good. Addressed in the next commit.

balajialg:
> Just as an aside, you can test the image on your own device with hubploy before waiting to see what issues shake out in CI. Once you've got hubploy, docker, etc. available locally, it is often faster than the change/deploy/CI cycle.

@ryanlovett Yes - going forward will try to install hubploy and test changes through that. Docker is already installed locally.

felder (changes requested):
The path in hubploy.yaml does not look correct.

.. in unix terms means "look in the directory above where I'm looking now"

Thus it's going to look for "deployments/image" which is not what you want.  Additionally if you compare hubploy.yaml for a11y with the hubploy.yaml for stat159 which you're modeling off of, you should see that "path" isn't defined at all.

Thus you likely just want to delete that line entirely from a11y/hubploy.yaml.

Issue #3942, opened 2022-11-14
Install python package allensdk for class MCB-160L


balajialg:
### Package Name

allensdk

### Hub details

- Hub: Datahub
- Language: Python

### Course Name

https://classes.berkeley.edu/content/2022-fall-mcellbi-160l-001-lec-001

### Semester Details

Fall 2022

### Installation Deadline

Students will use this package as part of this class on Thursday and Friday this week. Recommended instructor to use !pip install till the package gets added to the image.

felder:
The conda forge package for this requires python 3.6 and is also quite old.
https://anaconda.org/conda-forge/allensdk
vs
https://pypi.org/project/allensdk/

Having pip install it separately fails to compile.  Having pip install it from the conda environment downgrades a number of packages and produces package conflicts.

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
xarray-einstats 0.3.0 requires xarray>=0.16, but you have xarray 0.15.1 which is incompatible.
arviz 0.12.1 requires xarray>=0.16.1, but you have xarray 0.15.1 which is incompatible.
Successfully installed aiohttp-3.7.4 allensdk-2.13.6 argschema-3.0.4 async-timeout-3.0.1 boto3-1.17.21 botocore-1.20.112 cachetools-4.2.4 chardet-3.0.4 distro-1.8.0 glymur-0.8.19 hdmf-3.4.7 jmespath-0.10.0 marshmallow-3.19.0 matplotlib-3.4.2 ndx-events-0.2.0 psycopg2-binary-2.9.5 pynrrd-0.4.3 pynwb-2.2.0 requests-toolbelt-0.10.1 s3transfer-0.3.7 scikit-build-0.16.2 scikit-image-0.16.2 simpleitk-2.2.0 tables-3.6.1 xarray-0.15.1
```

Note that the allensdk docs also only guarantee compatibility with python 3.6 and 3.7, not 3.9 which is what is currently installed.

https://allensdk.readthedocs.io/en/latest/install.html

At this time I do not believe I can accommodate this request.

balajialg:
@petersudmant This class is affiliated with the Molecular and Cell Biology department. Is this something you could help us with by installing allensdk python package in the biology hub? Do you have some bandwidth?

petersudmant:
[...]

Issue #3939, opened 2022-11-11
Request python packages dbt-postgres, dbt-bigquery for class Data 101


lukeliu0415:
### Package Name

dbt-postgres, dbt-bigquery

### Hub details

- Hub: Data 101
- Language: Python

### Course Name

Data 101, Lakshya Jain
data101.org

### Semester Details

Fall 2022

### Installation Deadline

11/14. Thank you!

balajialg:
@lukeliu0415 Will these package versions work for you?

https://anaconda.org/conda-forge/dbt-postgres
https://anaconda.org/conda-forge/dbt-bigquery

Can we resolve this request by 11/15? Going forward, Would appreciate it if you give us at least 2 working days to resolve such requests.

lukeliu0415:
Hi! I've only tested out the installations with pip:
https://pypi.org/project/dbt-bigquery/
https://pypi.org/project/dbt-postgres/
If these are equivalent as the conda packages, they should work!

Thanks for letting me know about the request timeframe. Resolving before this Wed 11/16 would be great. Thanks!

felder:
This should be on prod in about 15m or so.

PR #3956, opened 2022-11-16
Adding packages.

felder:
Issues #3942 #3939


felder:
I surrender.

ericvd-ucb:
Bummer - what happened - some sort of loop of dependencies?  This is one of those things where this instructor has taught with this module material multiple times in the past!  - sorry just catching up on Slack now

Issue #3845, opened 2022-10-14
Adjust spawner messaging when servers are slow to start


ryanlovett:
# Summary

#3836 tracks an issue where servers were slow (~2m) to start up. The current messaging that students see is from kubespawner's [`progress`](https://github.com/jupyterhub/kubespawner/blob/71a9538a892d580476dd31d12ff17ed6217b8ea5/kubespawner/spawner.py#L2296) which emits events messages if available and if the feature is enabled.

Obviously we want to fix any underlying issues around slow server startup, but if they still occur for whatever reason, I think it'd be good to display something simple like "your server is slow to start up", and then something comforting and hopefully true like, "the hub may be busier than normal, please wait just a little while longer. (restarting your server will put you at the end of the queue.)"

Grafana/prometheus have access to real-world data on start up time, so a more advanced solution might involve the messaging displaying the recent startup time average or 90% percentile figure. Maybe that's a hub service which fetches the data from prometheus and exposes it via a simple api. Or maybe the spawner should query prometheus directly.

# User Stories

In this case, students had a bad experience because a 2m startup time seemed like infinity to them -- normally servers start up fairly quickly. It caused a lot of anxiety just before an assignment deadline and a gentle message could have gone a long way to comfort them. This is like when you call up a service over the phone and are put on hold. If you're told what your estimated wait time is, you will probably feel less frustrated.

# Acceptance criteria

- If server startup time exceeds some threshold, the hub's UI for spawner progress will display a message acknowledging the condition, possibly with some data on recent start up time.

balajialg:
@ryanlovett Thanks for raising this request! Seems super important. Is this UI change controlled by the Datahub team or part of the upstream?

ryanlovett:
@balajialg It is already configurable but requires a bit of dev (there’s no single bit to flip). I’m away from keyboard but I think hub lets you define the “slow server” threshold, and then you need to insert the messages.

ryanlovett:
Fixed by @shaneknapp !

PR #3963, opened 2022-11-18
add a timer and inject a message during slow server startup

shaneknapp:
implement a timer in the progress function to inject a message imploring for patience during slow server spinup.

https://github.com/berkeley-dsep-infra/datahub/issues/3845

### DIFF
--- a/hub/values.yaml
+++ b/hub/values.yaml
@@ -302,6 +302,14 @@ jupyterhub:
           """
         )
 
+        spawn_launcher_timer_frequency = Integer(
+          5,
+          config=True,
+          help="""
+          Display the patience message every 5 seconds.
+          """
+        )
+
         spawn_launcher_timer_message = Unicode(
           """
           Server launch is taking longer than expected. Please be patient!
@@ -488,7 +496,7 @@ jupyterhub:
               # patient
               # https://github.com/berkeley-dsep-infra/datahub/issues/3845
               if timer >= self.spawn_launcher_timer:
-                if timer % 10 == 0:
+                if timer % self.spawn_launcher_timer_frequency == 0:
                   # display only every 10 seconds
                   patience_message = textwrap.dedent(self.spawn_launcher_timer_message)
                   patience_message += " Current time spent waiting: %i seconds" % timer
--- a/hub/values.yaml
+++ b/hub/values.yaml
@@ -321,8 +321,6 @@ jupyterhub:
             return labels
 
           async def start(self):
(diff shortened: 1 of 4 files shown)


shaneknapp:
now...  how to test this?  XD

ryanlovett (commented):
Looks good! Minor suggestions: making `patience_message` a traitlet as well and inserting a comment at the top of `progress` that the code is mostly from KubeSpawner. (to deflect suggestions about alterations to *that* code and in case things need to be reconciled in the future)

shaneknapp:
> Looks good! Minor suggestions: making `patience_message` a traitlet as well and inserting a comment at the top of `progress` that the code is mostly from KubeSpawner. (to deflect suggestions about alterations to _that_ code and in case things need to be reconciled in the future)

yep, all sgtm++...

shaneknapp:
also general comment for the world:  converting the code from spawner.py (4 space tabs) to this YAML crap (2 space tabs) was an exercise in visually confirming that i was properly indenting.  and of course this can't be tested until i spin up my own hub and override start() with a delay...  i have VERY strong feeling about how this is done (python...  in yaml?  le sigh) and want to figure out a way to do this better.

hub/values.yaml
```diff
@@ -323,6 +325,8 @@ jupyterhub:
             return labels
 
           async def start(self):
+            # TODO: REMOVE BEFORE MERGING 3963 TO PROD!!!!!!!!!
+            sleep(45)
```
yuvipanda: You want `await asyncio.sleep(45)` here - otherwise the *entire process* will be blocked, and the code in the progress handler will not get a chance to execute until after this sleep call returns.

shaneknapp: ah, i was curious about that...  and it makes sense.  :)

hub/values.yaml
```diff
@@ -426,6 +446,84 @@ jupyterhub:
...
+
+              # check the timer, and if we're over 60sec ask the user to be
+              # patient
+              # https://github.com/berkeley-dsep-infra/datahub/issues/3845
+              if timer >= self.spawn_launcher_timer:
+                if timer % 10 == 0:
+                  # display only every 10 seconds
+                  patience_message = textwrap.dedent(self.spawn_launcher_timer_message)
+                  patience_message += "\nCurrent time spent waiting: %i seconds" % timer
+                  event["message"] = patience_message
```
shaneknapp: i can/should change `event["message"]` to something else

shaneknapp:
ok...  i've tested this pretty exhaustively and i think we should be g2g!

PR #3981, opened 2022-12-01
Adding apt-package "zip" as an explicit dependency in some of the hubs

### LINKED ISSUE #3979: "Download directory" option results in 5xx error messsages across Data 100, EECS and Stat 159 hubs
### Bug description

As a user, whenever I try to download directories in Data 100, Stat 159, or EECS hubs - I get the 500/503 error messages. It works well across other hubs like Public health, Data 8, Datahub, Data 101, Data 102, Biology hub, etc. This would be a useful feature for students to download their course data at the end of the semester. Sharing the error messages below,

image
image

### Environment & setup

- Hub:  Data 100, Stat 159, and EECS

### How to reproduce

1. Select a directory in Data 100, Stat 159, or EECS hub
2. Click on the "Download Directory" option
3. Wait for the 5xx error message

balajialg:
fixes #3979

### DIFF
--- a/deployments/astro/image/apt.txt
+++ b/deployments/astro/image/apt.txt
@@ -19,5 +19,5 @@ emacs
 # vim4ever!
 vim
 
-# https://github.com/berkeley-dsep-infra/datahub/issues/3979
+# For jupyter-tree-download. Ref: https://github.com/berkeley-dsep-infra/datahub/issues/3979			
 zip
--- a/deployments/astro/image/apt.txt
+++ b/deployments/astro/image/apt.txt
@@ -18,3 +18,6 @@ emacs
 
 # vim4ever!
 vim
+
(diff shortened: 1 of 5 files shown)


ryanlovett (commented):
Looks good! I'd also add it to `deployments/cee/image/apt.txt`. You might consider appending to the comment, "for jupyter-tree-download", or some such, since that is the extension that uses it. That way, if we ever want to remove the extension in the future, someone could find the comment with a `git grep jupyter-tree-download` and could consider removing zip as an unnecessary dependency.

balajialg:
@ryanlovett Thanks. I made a commit updating info about treedownload. Also, Added an entry for zip in the docker file for Datahub as a test. If it works then will replicate it across other images.

Issue #3983, opened 2022-12-03
2022-12-02 hub outage debrief!


balajialg:
We had an outage for 2 hours between 3 - 5 PM PST today which affected all users (mainly Data 8 considering that they had an assignment due today which had to get postponed by a day). Thanks to the efforts of @ryanlovett @yuvipanda and @felder we were able to get the hubs to be stable. We should debrief this issue next week extensively as we had a relatively stable last 3 weeks and made no major updates to the infrastructure in recent times. 

![image](https://user-images.githubusercontent.com/2306166/205421458-c72935d1-cac6-47d6-adc2-43a40dace42b.png)

We should write an incident report after doing an extensive analysis when we are back next week.

balajialg:
Drafted the outage and resolution communication process which needs to get followed whenever there is a major outage with all hubs - https://docs.google.com/document/d/1E32D-FAcFFpU5oRSzAAqfhUHMLvm3_K6Bv6nl9S2-DU/edit?usp=sharing. I am defining any major outage as "hubs not available for users for more than 30 minutes". If anyone has further inputs then please add your comments in the Google doc.

balajialg:
I will create a PR which has the template for incident report. @ryanlovett Can you review the information you shared in the Slack thread and add relevant information to the PR I am going to create in the next few days.

ryanlovett:
@balajialg Sure, just link to the PR in a comment here and I'll transfer info to it.

balajialg:
@ryanlovett FYI - Created a PR for incident report template - https://github.com/berkeley-dsep-infra/datahub/pull/4006. Let me know what your thoughts are!

Issue #3538, opened 2022-08-09
Improve Grafana dashboard's resilience!


balajialg:
### Bug description

Grafana dashboard threw a 503 error when there was an outage across all the hubs today. Considering that @felder uses grafana hub metrics to measure the health across all hubs, it is important that grafana gets isolated from such outages. 

Note: Grafana came back after all the hubs started operating. So, there is no issue with grafana as of now. 

![image (8)](https://user-images.githubusercontent.com/2306166/183542888-9ada0701-291c-41d3-99a4-0fe7e08c3a59.png)

Creating this issue to explore different design choices to avoid this scenario in the future. 

### Environment & setup

Grafana + Prometheus combination

### How to reproduce

See the above description!

balajialg:
One of the discussion items we had from a short meeting today was steps to improve Grafana's resilience. @shaneknapp had an idea that it would make sense to create a new monitoring node pool and move our monitoring infrastructure which includes Grafana and Prometheus to this node pool. Through this approach, we will still have the monitoring infra work when the hubs part of the core node pool goes down.

yuvipanda:
You should consider https://cloud.google.com/managed-prometheus too

balajialg:
Closing this issue as it is fixed by #3995.

Issue #3984, opened 2022-12-03
Use managed prometheus


ryanlovett:
# Summary

# User Stories

We currently deploy prometheus as part of our hub. This is fairly straightforward, however it becomes unavailable whenever something happens to the core nodes. For example today one of the nodes hit a very high load and, if we had an alert for such a condition, it wouldn't have been able to respond because prometheus was taken down by that high load.

Using the [managed service](https://cloud.google.com/stackdriver/docs/managed-prometheus) would avoid this.

# Acceptance criteria

- Use google's service rather than the in-cluster instance.

# Important information

- https://cloud.google.com/stackdriver/docs/managed-prometheus

# Tasks to complete

- bring up the managed service
- remove the in-cluster service

balajialg:
@ryanlovett Naive question, Looking at the historical data during the past seven days, it seems the load on Friday was not as high as the load during the previous two days. Do you have any hypotheses on why the load on Friday particularly had an impact?  Am I missing something by looking at this data?

![image](https://user-images.githubusercontent.com/2306166/205742715-b166daed-fd63-4242-98db-82efe2e4184b.png)
![image](https://user-images.githubusercontent.com/2306166/205742955-44115f64-63ce-4f15-95cb-fe5a79adb73f.png)

ryanlovett:
@balajialg I would guess it has something to do with the academic cycle since it is the Friday before RRR week. You could drill down and see if one particular hub had much less usage or if there was a drop across all of them.

balajialg:
@ryanlovett It was a drop across almost all the major hubs. Ref:  https://docs.google.com/document/d/1hw3wR_1Dc40pm7OsZYubzrkk6SD4q3vzA6i89TqStKE/edit?usp=sharing. Getting more curious why we had an outage despite having traffic that was lower than the previous days across multiple hubs.

yuvipanda:
So, I created a new nodepool for support, and gave it 48G of RAM, and tried to get list of running users over last 90 days.

It failed with a timeout, and actual memory usage never went past 8G.

So went back, and looked at the PromQL query itself:

```
# Sum up all running user pods by namespace
sum(
  # Grab a list of all running pods.
  # The group aggregator always returns "1" for the number of times each
  # unique label appears in the time series. This is desirable for this
  # use case because we're merely identifying running pods by name,
  # not how many times they might be running.
  group(
    kube_pod_status_phase{phase="Running"}
  ) by (pod)
[...]

PR #3995, opened 2022-12-08
Move all support stuff except nginx to support pool

yuvipanda:
Not moving nginx as that was the cause (perhaps?) of
outage in https://github.com/berkeley-dsep-infra/datahub/issues/3983.

Ref https://github.com/berkeley-dsep-infra/datahub/issues/3538


yuvipanda:
Note that I don't think this will actually solve any of the prometheus issues - see https://github.com/berkeley-dsep-infra/datahub/issues/3984#issuecomment-1343125440

PR #4017, opened 2022-12-20
Sync repo2docker between build and deploy.

ryanlovett:
... and the other instances in singleuser images. Its inclusion in requirements.txt keeps them the same. This also pins it at something newer. We also need to explicitly set python to 3.9 in repo2docker builds, otherwise they use python 3.7. Consider bumping this in a separate PR.


ryanlovett:
Relevant: https://github.com/jupyterhub/repo2docker/pull/1219

PR #4002, opened 2022-12-13
filestore testing do not merge me

felder:

Issue #3898, opened 2022-10-31
Explore Google File Store as a replacement for NFS!


balajialg:
We have been facing NFS issues due to a race condition in the Linux kernel which is hard to troubleshoot and resulted in a few outages in the past month. During our last team meeting, some of us were interested to explore Google FIle Store (GFS) as an alternative to NFS. @ericvd-ucb kindly agreed to do the outreach and reached out to Google Filestore folks using his contacts to have a conversation about GFS instead of NFS. GFS's point of contact outlined the below points in their response,

_What you described is common when customers run NFS themselves. These consistency issues are hard to troubleshoot. We do have many customers running Filestore with multiple directories that in turn serve multiple users. The benefit of using a managed NFS solution like Filestore is that you don't have to manage NFS and simply get it out of the box. Filestore has [multiple tiers ](https://cloud.google.com/filestore/docs/service-tiers)(we recommended Filestore Enterprise to give you a HA solution by default) but you can also choose basic if you like. You only pay for the storage you consume. (as opposed to when you are running NFS yourself you probably are consuming compute from VMs and storage form PDs) The one thing to watch out for with Filestore Enterprise (based on what we hear from other customers) is the entry point of 1 TiB. You can of course consume the space by placing the directories of multiple users in the same Filestore instance, driving up utilization. In case you want isolation between users, you can also use [multishares](https://cloud.google.com/filestore/docs/multishares) that share the underlying Filestore instance and drive up utilization. Outside this specific concern of the min entry size (that you can workaround based on solutions shared above), you get regional-backed storage, managed NFS, and just pay for the storage consumed and many customers use it at scale._

**We need to evaluate whether what they proposed above is something we are interested to explore from a technical standpoint.** 

From my limited understanding, I looked at our billing report for the month of Oct 22 and found that their enterprise version  (~$600 per month for 10 TiB) is at par with whatever we are spending for PD + snapshots (~4100 per month for 70 TiB). I am assuming I didn't miss anything in this calculation but please correct me if my interpretation is wrong.

![image](https://user-images.githubusercontent.com/2306166/199128242-82498eaf-89af-4553-80e9-8a37a647c584.png)

# To Do
- [x] Evaluate whether GFS fits our need
- [x] Deploy GFS on a pilot hub
- [x] If the pilot is successful, transition all hubs from NFS to GFS

ryanlovett:
Thanks for looking into this @balajialg !

[...]

PR #4059, opened 2022-12-29
initial commit of GFS squash flags (do not merge)

shaneknapp:
initial commit of GFS squash configs for each deployment.

after setting up a bunch of GFS instances, i've discovered that it's much safer to have a deployment-specific config in each dir.

it's also important that the instance size in these files matches what's reported in gcloud/google cloud console.  if the number in here is lower, it will fail.  if it's higher, it will **quietly scale** the instance up!

https://github.com/berkeley-dsep-infra/datahub/issues/3898
https://github.com/berkeley-dsep-infra/datahub/pull/4002


shaneknapp:
idea:  move these files to `config/filestore`, and stick a README.md in each dir w/the three-step instructions on how to deploy.  thoughts?

ryanlovett:
I think the content of the READMEs can probably be moved to the docs somewhere, but it makes sense for the json to be in each deployments config.

I think it also makes sense to put the json into the cookiecutter template and to prompt for the storage size.

docs/admins/howto/new-hub.rst
```diff
@@ -11,9 +11,8 @@ Why create a new hub?
 
 The major reasons for making a new hub are:
 
+#. A new course wants to join the Berkeley Datahub community!
 #. You want to use a different kind of authenticator.
-#. Some of your *students* are *admins* on another hub,
```
ryanlovett, line 15: Isn't this still relevant?

ryanlovett (commented):
This looks good to me, and thanks for adding the cookiecutter template and docs update!

Besides the doc change about student/admin access, just one other comment. I think the hub name doesn't need to be part of the squash-flags filename, e.g. it could just be `{hub}/config/filestore/squash-flags.json` for each, which is analogous to the various config/secret files like `common.yaml`, `prod.yaml`, `gcr-key.json`, etc. If you agree then the docs mentioning this file would need to be changed as well.

balajialg:
Thanks for the docs update @shaneknapp!

shaneknapp:
> This looks good to me, and thanks for adding the cookiecutter template and docs update!
> 
> Besides the doc change about student/admin access, just one other comment. I think the hub name doesn't need to be part of the squash-flags filename, e.g. it could just be `{hub}/config/filestore/squash-flags.json` for each, which is analogous to the various config/secret files like `common.yaml`, `prod.yaml`, `gcr-key.json`, etc. If you agree then the docs mentioning this file would need to be changed as well.

the primary driver for explicitly naming these files is to avoid confusion and accidentally upscaling storage when modifying `ROOT_SQUASH`.  granted, we probably won't be doing work en masse like this again (or for a long time), so it's probably fine to just name things `squash-flags.json`.

docs/admins/howto/new-hub.rst
```diff
@@ -54,25 +64,40 @@ we provide that sets up a blank hub that can be customized.
...
+
+   .. code:: bash
+
+         drwxr-xr-x 4 ubuntu ubuntu     45 Nov  3 20:33 a11y-filestore
+	 drwxr-xr-x 4 ubuntu ubuntu     33 Jan  4  2022 astro-filestore
+	 drwxr-xr-x 4 ubuntu ubuntu  16384 Aug 16 18:45 biology-filestore
+
+   #. Now we need to set ``ROOT_SQUASH`` on the newly created mount. In the 
+      ``datahub/deployments/<hubname>/config/filestore/`` directory is a file named 
+      ``<hub-name>-squash-flags.json``. You will use this file when running the following ``gcloud``
```
ryanlovett: With the latest changes you can remove `<hub-name>-` from `<hub-name>-squash-flags.json`.

docs/admins/howto/new-hub.rst
```diff
@@ -54,25 +64,40 @@ we provide that sets up a blank hub that can be customized.
...
+	 drwxr-xr-x 4 ubuntu ubuntu  16384 Aug 16 18:45 biology-filestore
+
+   #. Now we need to set ``ROOT_SQUASH`` on the newly created mount. In the 
+      ``datahub/deployments/<hubname>/config/filestore/`` directory is a file named 
+      ``<hub-name>-squash-flags.json``. You will use this file when running the following ``gcloud`` 
+      command to apply the changes:
+
+   ..code:: bash
+
+     gcloud filestore instances update <filestore-instance-name> --zone=us-central1-b --flags-file=<hubname>-squash-flags.json
```
ryanlovett: Similar to the one above, you can remove `<hubname>-` from `<hubname>-squash-flags.json`.

ryanlovett (approved):
LGTM! Just two doc changes which correspond to the filename change.

PR #4066, opened 2023-01-04
Changing node pools and storage locations for r, datahub, data8, biology, and data100

felder:
Also see PR #4002

### DIFF
--- a/deployments/biology/config/common.yaml
+++ b/deployments/biology/config/common.yaml
@@ -1,7 +1,7 @@
 nfsPVC:
   enabled: true
   nfs:
-    serverIP: nfsserver-01
+    serverIP: 10.120.80.242
 
 jupyterhub:
   scheduling:
@@ -56,11 +56,11 @@ jupyterhub:
 
 
     nodeSelector:
-      hub.jupyter.org/pool-name: beta-pool
+      hub.jupyter.org/pool-name: biology-pool
     storage:
       type: static
       static:
-        pvcName: home-nfs
+        pvcName: home-nfs-v3
         subPath: "{username}"
       extraVolumeMounts:
         - name: home
--- a/deployments/biology/config/prod.yaml
+++ b/deployments/biology/config/prod.yaml
@@ -1,6 +1,6 @@
 nfsPVC:
   nfs:
-    shareName: export/homedirs-other-2020-07-29/biology/prod
+    shareName: shares/biology/prod
 
 jupyterhub:
   ingress:
--- a/deployments/biology/config/staging.yaml
+++ b/deployments/biology/config/staging.yaml
@@ -1,6 +1,6 @@
 nfsPVC:
   nfs:
-    shareName: export/homedirs-other-2020-07-29/biology/staging
+    shareName: shares/biology/staging
 
 jupyterhub:
   scheduling:
--- a/deployments/data100/config/common.yaml
+++ b/deployments/data100/config/common.yaml
@@ -1,7 +1,7 @@
 nfsPVC:
   enabled: true
   nfs:
-    serverIP: nfsserver-01
+    serverIP: 10.203.15.114
 
 jupyterhub:
   hub:
@@ -46,7 +46,7 @@ jupyterhub:
     storage:
       type: static
       static:
-        pvcName: home-nfs
+        pvcName: home-nfs-v3
         subPath: "{username}"
       extraVolumeMounts:
         - name: home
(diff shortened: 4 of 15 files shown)


felder:
Rotating node pools and storage for datahub:

When using the console to add a node pool, I noticed that there is an option to get the command line for pool creation.  By comparing the options for creating a new node pool with an already existing one in the console, I was able to obtain the command line below.  I then grouped the options into multiple lines that I believe make the most sense for us with the goal of placing options that need to be changed near the beginning of the command.  At a minimum you'll want to change the name of the pool (after node-pools create) as well as the node-labels (hub.jupyter.org/pool-name=datahub-pool).  

The options in the first four lines should be pretty self explanitory.  The fifth line contains metadata added by us which I assume must be there.  In particular "--tags hub-cluster" is how firewall access is managed.

```
gcloud beta container node-pools create "user-datahub-2023-01-04"  \
  --node-labels hub.jupyter.org/pool-name=datahub-pool --machine-type "n2-highmem-8"  \
  --enable-autoscaling --min-nodes "0" --max-nodes "3" \
  --project "ucb-datahub-2018" --cluster "fall-2019" --region "us-central1" --node-locations "us-central1-b" \
  --node-taints hub.jupyter.org_dedicated=user:NoSchedule --tags hub-cluster \
  --image-type "COS_CONTAINERD" --disk-type "pd-balanced" --disk-size "200"  \
  --metadata disable-legacy-endpoints=true \
  --scopes "https://www.googleapis.com/auth/devstorage.read_only","https://www.googleapis.com/auth/logging.write","https://www.googleapis.com/auth/monitoring","https://www.googleapis.com/auth/servicecontrol","https://www.googleapis.com/auth/service.management.readonly","https://www.googleapis.com/auth/trace.append" \
  --no-enable-autoupgrade --enable-autorepair --max-surge-upgrade 1 --max-unavailable-upgrade 0 --max-pods-per-node "110"
```
  
Once the new node pool is created, you need to modify common.yaml for the deployment.  In this case:
`vi deployments/datahub/config/common.yaml`

Alter jupyterhub->singleuser->nodeSelector and change it from alpha-pool to datahub-pool.

Additionally if you're moving the hub to new storage (google filestore in this case):

1.  Create a new filestore instance.
2.  Do an initial data sync
3.  Grab the IP for it the filestore instance (I used mounts on the nfsserver)

4.  modify common.yaml for the deployment.  In this case:
`vi deployments/datahub/config/common.yaml`
Alter jupyterhub->singleuser->storage->static->pvcName and change it from home-nfs to home-nfs-v3

5.  modify staging.yaml for the deployment.  In this case:
`vi deployments/datahub/config/staging.yaml`
Alter nfsPVC->nfs->shareName and change it to shares/datahub/staging

6.  modify prod.yaml for the deployment.  In this case:
`vi deployments/datahub/config/prod.yaml`
Alter nfsPVC->nfs->shareName and change it to shares/datahub/prod

Now you need to delete all the user pods and prevent new users from signing in.  To do that:

1.  Prevent new users by stopping the hub pod for staging and/or prod:
```
kubectl -n datahub-staging scale --replicas=0 deployment/hub
kubectl -n datahub-prod scale --replicas=0 deployment/hub
```
2.  Delete the user pods (there's probably a better way to do this):
```
kubectl -n datahub-staging get pods | grep jupyter | cut -f1 -d ' ' | xargs -n1 -I {} kubectl -n datahub-staging delete pods {} --wait=false
kubectl -n datahub-prod get pods | grep jupyter | cut -f1 -d ' ' | xargs -n1 -I {} kubectl -n datahub-prod delete pods {} --wait=false
```

3.  Confirm user pods are gone:
```
kubectl -n datahub-staging get pods
kubectl -n datahub-prod get pods
```

Do any final storage syncs you need to do.

Lastly if you're changing storage, before deployment you need to delete the physical volume claims (pvc) and physical volumes (pv) otherwise helm will error:
```
kubectl -n datahub-staging delete pvc home-nfs
kubectl -n datahub-staging delete pvc home-nfs-v3
kubectl delete pv datahub-staging-home-nfs
kubectl delete pv datahub-staging-home-nfs-v3

kubectl -n datahub-prod delete pvc home-nfs
kubectl -n datahub-prod delete pvc home-nfs-v3
kubectl delete pv datahub-prod-home-nfs
kubectl delete pv datahub-prod-home-nfs-v3
```
Now deploy.  When you deploy new pv and pvc should be created.  Additionally the hub pods will be restarted.

shaneknapp (approved):

ryanlovett (approved):
Looks good to me! Only suggestion would be to turn your detailed comment into a new doc. (either this PR or a new one)

balajialg (approved):

PR #4075, opened 2023-01-06
migrate dlab/prob140 to new nodepools/gfs

shaneknapp:


shaneknapp:
these have also been deployed via helm, and i'll merge to staging now

Issue #4057, opened 2022-12-28
Request more RAM for class ARE 212


ligon:
**Which hub do you want more RAM on?**

datahub

**Which class is this request for?**

Spring 2023, ARE212, "Econometrics: Multiple Equation Estimation"

**What is the Academic Guide URL for this class?**

https://classes.berkeley.edu/content/2023-spring-aresec-212-001-lec-001

**What is the bCourses URL for this class?**

https://bcourses.berkeley.edu/courses/1521412

**How many students do you expect in this class?**

20

**How much RAM per user is needed?**

4GB

**Why does this class need this much RAM?**

Students are asked to tackle estimation problems using pandas DataFrames that use this much RAM.

**By what date (MM/DD) do you want this request to be fulfilled?**

Before start of Spring Semester!

**Any additional information we should know about?**

Thanks for providing this amazing service!

balajialg:
@ligon Both the requests raised by you to increase resources are merged into production. Can you please test it and see if it works? I am closing your issues for now but please feel free to reopen them if you face any issues. Thanks!

ligon:
When I log into datahub I see the requested amount of RAM.  

This does not test the experience a student enrolled in one of those two classes would have, of course, and I don't know how to test this (aside from asking students in the two classes to check).

I'll do that next week unless there's a better way.

balajialg:
@ligon Thanks for checking! Yes - I think the best way to test this would be to ask students to check it when they log in to datahub. If any of them have issues then please do report them back to us!

ligon:
At least some students in the class do **not** have the additional requested RAM (they have the default 1G).

We have gotten to the point in the semester where they really need the additional RAM!   Help!

PR #4080, opened 2023-01-06
Increase RAM for courses ENVECON 153 and ARE 21

### LINKED ISSUE #3982: Request more RAM for ENVECON 153
**Which hub do you want more RAM on?**

datahub

**Which class is this request for?**

https://classes.berkeley.edu/content/2023-spring-envecon-153-001-lec-001

**How many students do you expect in this class?**
54 students

**How much RAM does this class need?**

3GB

**Why does this class need this much RAM?**

Students analyze some datasets which require that much ram.

**By when (MM/DD) do you want this request to be fulfilled?**

Start of Spring semester

**Any additional information we should know about?**

Thank you!  Love teaching with datahub!

balajialg:
fixes #3982 and #4057

### DIFF
--- a/deployments/datahub/config/common.yaml
+++ b/deployments/datahub/config/common.yaml
@@ -130,3 +130,12 @@ jupyterhub:
       1519998: # Legal Studies 123, Spring 2023, issue #3718
         mem_limit: 3072M
         mem_guarantee: 3072M
+      
+      1521830: # ENVECON 153, Spring 2023, issue #3982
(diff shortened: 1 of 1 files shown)


balajialg:
@ryanlovett Can you please merge this if my config values are accurate?

ryanlovett:
The configuration is syntactically correct, and thanks for putting the course info in the comments!

I've posted a follow up question in slack, re: guarantee == limit.

PR #4160, opened 2023-01-24
Add nbgitpuller 

fperez:
I'd accidentally removed it a few days ago - it turns out we do need it.


fperez:
Sorry @shaneknapp  - I _thought_ we weren't using nbgitpuller but in fact we do for our labs, so we should put it back in our image. My bad :)

fperez:
Thanks @shaneknapp!

PR #4232, opened 2023-02-09
Merge RTC-related PRs.

ryanlovett:
@fperez gave the go ahead in person, after discussion about the link sharing authorization issue that @balajialg identified. Collaborators will not be able to start another user's server, and still require a specific link shared by the owner and unique to a server's lifetime.

### DIFF
--- a/deployments/stat159/config/common.yaml
+++ b/deployments/stat159/config/common.yaml
@@ -38,6 +38,8 @@ jupyterhub:
       # Tell code where to display GUIs
       # The VNC /desktop link must be opened already for this to work
       DISPLAY: ":1.0"
+      # Unset NotebookApp from hub/values. Necessary for recent lab versions.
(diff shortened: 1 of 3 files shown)


fperez:
Awesome, many thanks @ryanlovett! 

Quick q - do you have any summary docs of these deployments? It might be nice to copy down into a single page a quick summary of:

1. The basic setup needed for this to work (packages/versions needed, modifications to startup config/flags).
2. A summary of the security implications, with links to the various issues we have open tracking this.

Nothing huge (not worth writing a ton yet as it's a rapidly changing topic) but useful to have a short summary of the key points in one place.

ryanlovett:
@fperez We don't have any descriptive text about specific deployments, just the build materials in the deployments/{deployment}/image/ directory. I wonder if we should create material under docs/ that describe the purpose of each hub, and related matters like the security implications you mention.

I can create a new issue for that.

Issue #4237, opened 2023-02-09
test/fix prometheus/grafana server start time metric by adding a delay to notebook at server launch


shaneknapp:
grafana is telling us that our server start times are never longer than 2 minutes.  anecdotal, and my actual experience, show that to be false (i had a server start after 3m, and checked grafana and it showed 2 minutes max).

<img width="483" alt="LIES" src="https://user-images.githubusercontent.com/1606572/217946469-7c469a87-6eb3-4fee-a39f-116258752e88.png">

so, i propose adding a traitlet to delay server start.  we can then set the timer to 5 minutes, hubploy to a11y-staging, and confirm that grafana lies.

after that, figure out what's broken and fix it.

shaneknapp:
this will help us debug and improve performance in cases like [this](https://github.com/berkeley-dsep-infra/datahub/issues/4009#issuecomment-1423335899)

felder:
omg...

@shaneknapp Is it possible there is a delay between server start and notebook launch?  We saw that in the meeting this afternoon when our student worker was launching a notebook in eecs.  If there is, that could also explain this behavior.

balajialg:
Hey, @shaneknapp For the #4009 issue you referenced - How did you come to the conclusion that the calendar events not spinning up enough nodes before the classes is no where related to the slow startup times? Based on the below data?

![image](https://user-images.githubusercontent.com/2306166/217982764-d89d255a-ca98-462e-8d9f-5b50ffefc430.png)

![image](https://user-images.githubusercontent.com/2306166/217982951-5eeba8a0-d4f1-4ba6-bc70-cb134ceebaab.png)

shaneknapp:
> omg...
> 
> @shaneknapp Is it possible there is a delay between server start and notebook launch? We saw that in the meeting this afternoon when our student worker was launching a notebook in eecs. If there is, that could also explain this behavior.

yes, it's entirely possible.

however, i just launched a server on a11y-staging w/a 3m delay, and grafana claims that it took 2m.  :\

![image](https://user-images.githubusercontent.com/1606572/217999983-9ca5402e-6ed1-4467-ba38-4238a723e0e2.png)

so:

1) grafana is lying, maybe because
2) prometheus is lying, or
3) we're querying the wrong metric for Server Start time, or
4) the underlying metric that we need is teh borked, or
5) all of the above.

shaneknapp:
[...]

Issue #4009, opened 2022-12-19
Increase compute resources for Course PS 3 January - April 2023 Mondays/Wednesdays 5-6:30 pm


dbroockman:
### Course Name

PS 3

### Detailed Requirements

350 students will be using R.Datahub simultaneously from 5-6:30 pm on Mondays and Wednesdays during spring 2023

### Semester Details

No

### Request Deadline

First class January 23
Last class April 26

balajialg:
@dbroockman We are making major upgrades to the infrastructure which might affect the current functionality of the calendar-based auto scaler. We are looking at a 2-3 week timeline to make these changes. We should be able to look into this request and resolve it once the upgrades are completed.

balajialg:
@shaneknapp Can we increase resources for this class? or do we need to tweak the script for auto scaler further?

balajialg:
@felder @shaneknapp @ryanlovett Created a recurring calendar event for Pol Sci 3 between 4.30 - 6.30 PM every monday and wednesday and allocated 8 nodes for this class of 350 students (~45 pods per node amounting to ~8 nodes) in the R hub. Let me know if you think the allocation makes sense?

ryanlovett:
@balajialg That number should not be set to how many total nodes should be allocated. It is how many available spares ("hotspares" or "placeholders") there should be. It is to address the rate of growth so if a lot of people login in a short amount of time then the hot spare value is higher. If all 350 login at exactly 4:30p, try setting the value to be 2 starting at 4:20p. The cluster will make two additional empty nodes available which can handle 90 logins (based on your pod density figure), not counting those that are already running user pods. The cluster will continue to ramp up nodes, always keeping "+2" empty. Once class is underway and the rate of growth has slowed, the hotspare/placeholder count can be set back to 1 or 0.

If 2 proves to be insufficient it can be scaled up higher, or if it is more than enough it can be set lower. You can verify by seeing how many users are in pending and how long server startup takes.

balajialg:
Thanks, @ryanlovett for the detailed explanation! It makes complete sense. I misunderstood that google calendar-based auto-scaler events allocate the exact nodes highlighted in the calendar event. This belief got strengthened after I saw that 10 hot spares got allocated for this class during Spring 2022 (Snapshot attached).  Changed the calendar entry to two nodes for now and let's tweak it based on real-time metrics (as you highlighted).

[...]

PR #4241, opened 2023-02-09
add server spawn delay

shaneknapp:
see https://github.com/berkeley-dsep-infra/datahub/issues/4237 and https://github.com/berkeley-dsep-infra/datahub/issues/4009#issuecomment-1423335899

### DIFF
--- a/hub/values.yaml
+++ b/hub/values.yaml
@@ -280,6 +280,15 @@ jupyterhub:
           return (x > y) - (x < y)
 
         class CustomAttrSpawner(KubeSpawner):
+          spawn_launcher_delay = Int(
+            0,
+            config=True
+            help="""
+            Time in seconds to delay a server launch, useful for debugging.  If set to zero,
+            no delay will take place.
+            """
+          )
+
           spawn_launcher_timer = Int(
             60,
             config=True,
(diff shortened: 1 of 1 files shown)


felder (commented):
Were you able to test this in staging via hubploy first?

hub/values.yaml
```diff
@@ -328,6 +337,10 @@ jupyterhub:
               self.mem_guarantee = override['mem_guarantee']
 
           async def start(self):
+            # delay if testing, see https://github.com/berkeley-dsep-infra/datahub/issues/4237
+            if self.spawn_launcher_delay:
```
ryanlovett, line 341: Will this always be True if the default is `180` ? Should the default be `0` for that reason?

I might not know the behavior of traitlets well enough -- maybe if the admin doesn't actually set the var, then the `if self.spawn_launcher_delay` will fail. Based on the [docs](https://traitlets.readthedocs.io/en/stable/config.html#python-configuration-files), I think the default value will be used.

So basically I think the default value should be 0, and the `if` should check for non-zero.

shaneknapp: default will indeed be 0.  i set it to 180 for testing purposes.

shaneknapp:
> Were you able to test this in staging via hubploy first?

yep, only a11y-staging has this new "feature" deployed.

hub/values.yaml
shaneknapp: and i pushed a new commit setting this back to 0

shaneknapp:
ok, this works and has already shown itself to be useful!  will be merging to staging now.

Issue #4251, opened 2023-02-13
Request python package sqlalchemy downgrade to 1.4.46 for Econ 148


oooyiyangc:
### Package Name

sqlalchemy==1.4.46 (downgrade from current 2.0.1)

### Hub details

- Hub: Datahub
- Language: Python

### Course Name

[https://classes.berkeley.edu/content/2023-spring-econ-148-001-lec-001](https://classes.berkeley.edu/content/2023-spring-econ-148-001-lec-001)
Econ 148: Data Science for Economists 
Instructor: Prof Eric Van Dusen

### Semester Details

Spring 2023

### Installation Deadline

Hopefully by 02/20 (next Monday) if possible. Sorry for the very short notice!

oooyiyangc:
@ericvd-ucb

balajialg:
Done

PR #4264, opened 2023-02-15
added sqlalchemy to requirements.txt

rohanjha123:
Doing this to ensure data 100 doesn't run into the same issue as https://github.com/berkeley-dsep-infra/datahub/issues/4251


ryanlovett:
Hi @rohanjha123, it looks like Mac-specific metadata files crept in. (".DS_Store")

rohanjha123:
Thanks for the catch, removed all the DS_Stores!

PR #4274, opened 2023-02-17
Update NodeJS for better MyST support

fperez:
This PR updates NodeJS to v18 so we can install the MyST CLI tools, and also updates the JupyterLab plugin (which changed names and is now at v 1.0 under the `jupyterlab_myst` package, which will be its new long-term home.


rowanc1:
🚀

fperez:
Thx @ryanlovett!!

PR #4278, opened 2023-02-18
Update TeX packages and clean up unnecessary repo2docker

fperez:


fperez:
Huge thanks @ryanlovett!! Just tested it on staging, all looks solid. Good to merge any time into prod (but no rush).

PR #4288, opened 2023-02-22
Move fileid database to /tmp/.

ryanlovett:
As pointed out by @yuvipanda in 2i2c-org/infrastructure#2246. We don't need the YAML anchor right now, but I'm preserving it in case it is useful later.

### DIFF
--- a/deployments/stat159/config/common.yaml
+++ b/deployments/stat159/config/common.yaml
@@ -51,6 +51,16 @@ jupyterhub:
           # https://github.com/berkeley-dsep-infra/datahub/issues/3160
           ContentsManager:
             allow_hidden: true
+      jupyter_server_config.json:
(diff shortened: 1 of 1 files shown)


yuvipanda:
https://github.com/jupyter-server/jupyter_server_fileid/issues/60 is the upstream issue

ryanlovett:
Thanks @yuvipanda ! I'm currently waiting on merging in case it is necessary to preserve the db across server sessions. (as you asked about)

ryanlovett:
Based on https://github.com/jupyter-server/jupyter_server_fileid/issues/60#issuecomment-1440932278,
> If RTC is persisting the fileids in their own persistent store (which I believe they do, but am not certain of this), then the fileid database also must be persisted. Otherwise, no.

Since we're not persisting the RTC db, it sounds like we don't need to persist this one. However down the road there may be other extensions that make use of the fileid.

PR #4335, opened 2023-03-07
Tex live bump

d-alex-hughes:
This PR alters how LaTeX is managed. To this point, we've been managing using `tlmgr` on an annually frozen TeX distribution. 

- The current build was frozen to a 2021 vintage; 
- The current TexLive vintage is a 2022 vintage; 
- There is a new vintage released every year, and so we'd have to maintain updates annually. 

The result of falling behind is a warning presented to students; and a ~5-10 minute install on student sessions. This feels like a poor allocation of central resources (i.e. to build for all students) and student time (i.e. 10m install). 

The PR pulls management from `tlmgr` in the Dockerfile, and makes two additions: 

1. It adds the R package `tinytex` [link](https://yihui.org/tinytex/) which is provided by a very reputable source. This is probably the preferred mode across most R users. This tool would be used by the student/user if they need to tinker with their install. 
2. It adds the TinyTex install using a `wget` call. This command line install is recommended by the author [here](https://yihui.org/tinytex/#installation). 

This change *could* be breaking -- we are removing the install of a large TeX install. It should **not** be breaking for a user who is working from the R/Rstudio side; it **could** (although I have not tested) be breaking for people who are using TeX from the jupyter side of a hub. 

A future PR might also move to install quarto [link](https://quarto.org), which could provide a unified TeX distribution.


ryanlovett:
@d-alex-hughes Fwiw, when we ran into the issue of students having to wait to install various TeX packages on other hubs, we [installed latex](https://github.com/berkeley-dsep-infra/datahub/blob/staging/deployments/stat20/image/Dockerfile#L29) from upstream Ubuntu packages. There haven't been any problems with it.

d-alex-hughes:
This PR implements the TeX management proposed by @ryanlovett and implmented in the Stat20 deployment. It also makes two additional changes. 

1. A small change, that we can revert if you like: Moves to enforce alphabetization ordering in the apt-get section. There's one orphan install of a fonts package that is far away from the rest of the TeX packages, but that's the only thing that seems lost. 
2. A larger change to include R package installs for all packages that are taught in the School of Information stats sequence. This now covers three coures, MIDS 203, MIDS 241, MIDS 271 so that a student has a fully working stack from the beginning. 

# Checks 

- By-hand checks are all "working" 
- The CircleCI checks have a lot of failures, but they're reading like rate limit failures, not actual failures. 

I think that this is ready to merge to staging, and would close #4335 if so.

ryanlovett:
This PR is cursed with some sort of github API issue at CircleCI. I'm closing in favor of #4355 which got past the issue.

d-alex-hughes:
@ryanlovett, after working back and forth with you for the datahub I talked some with folks over at Rocker (also here on campus) to see their thoughts on the issue. They came up against this once previously, in 2021, and have an idea out now to work with TexLive maintainers to add a check to suppress this warning when we are in the freeze period. 

https://github.com/rocker-org/rocker-versioned2/issues/609

ryanlovett:
Interesting, thanks! Yeah, it was Carl in another related issue who recommended installing TeX from the distro.

Issue #4313, opened 2023-03-02
Plan for hubs that are not used this semester!


balajialg:
# Summary

Last week's investigation revealed that Data 101 hub had postgres and mongo db PVC's mounted even while not being used this semester. I reached out to the Fall 22 Data 101 instructor Lakshya Jain who confirmed that he is not teaching this semester using this hub. Lisa Yan who plans to teach Data 101 during FA 23 confirmed that she plans to use the hubs **starting in August.** 

We need to discuss and decide on a policy for handling instances where the deployed hubs are not actively used during a semester. We are spending on monthly filestore and node pools for this hub which could exceed thousands of dollars by July 23. 

We should either a) Decommission unused hubs and create new ones based on instructor request or b) Either delete their filestore instances or move them to shared filestores, delete their node pools or add them to common pools, and archive user files stored in their home directories.

I will use this GitHub issue to track all the hubs that are not used during SP 23.

Hubs | Hub Status | Confirmed by | Additional Notes | Decision Made | Decision Communicated to Instructor
-- | -- | -- | -- | -- | --
Data 101 Hub | Not used | Lakshya Jain (FA 22 instructor for Data 101) and Lisa Yan (FA 23 instructor for Data 101) | She will use the hub starting August | Students are given a week's time to download their data | Instructors gave a thumbs up for deleting student data  
Astro Hub | Not used |  Josh Bloom (FA 22 instructor for Astro) | He will definitely use it for FA 23 and may use it during Summer 23 | Delete student data and explore folding Astro hub into main Datahub |Gave a thumbs up for deleting student data and requested a plan of action for handling package conflicts
CEE Hub | Not used|  Nicholas Sitar (FA 22 instructor for Civ Eng 170) | He will definitely use it for FA 23 and is on sabbatical during SP 23 | Delete student data | Instructor gave a thumbs up and informed students to back up.

# Tasks to complete

- [x] Decide policy on unused hubs
- [x] @balajialg to do outreach to instructors using Astro and CEE hub and update this list
- [x] Decommission Astro Hub
- [x] Decommission CEE Hub
- [x] Decommission Data 101 Hub

balajialg:
- [x] Figure out UI for folks accessing these hubs?
- [x] Make a decision: Delete hubs? what happens to user files?

shaneknapp:
filestore:  we can create backups, delete the instance, and create a new instance from the backup

backups are $.08/G/month

[...]

PR #4340, opened 2023-03-08
set node placeholders to zero for CEE, astro and data101 hubs

shaneknapp:
https://github.com/berkeley-dsep-infra/datahub/issues/4313


ryanlovett:
@shaneknapp This looks a bit different than most chartpress changes. The tags here use `0.dev.git` instead of the `n1234` syntax, and a tarball was added to hub/charts which we don't normally have. See https://github.com/berkeley-dsep-infra/datahub/tree/staging/hub.

Maybe the tagging difference is due to your version of chartpress, in which case it'd probably be fine?

```shell
% chartpress --version
chartpress version 1.3.0
```

I've never added anything to hub/charts/ though.

ryanlovett:
Ahh, see https://github.com/jupyterhub/chartpress/blob/main/CHANGELOG.md#breaking-changes.

PR #4350, opened 2023-03-09
adding a gitignore

shaneknapp:
coz i'm surprised we don't already have one!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,11 @@
+# hubploy, chartpress
+docs/_build
+hub/charts
 **/*.lock
 **/*.tgz
+
+# vi
 **/*.swp
+
+# python
 **/__pycache__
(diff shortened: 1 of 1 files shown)


ryanlovett:
Looks good! Some others worth ignoring:
 - `docs/_build` - for those who test docs builds
 - `hub/charts` - for those who run chartpress

Regarding vi swap files, I put them in my own ~/.gitignore via `git config --global core.excludesfile ~/.gitignore`.

shaneknapp:
> Regarding vi swap files, I put them in my own ~/.gitignore via `git config --global core.excludesfile ~/.gitignore`.

yeah, but not everyone does it that way...  doesn't really matter to me either way, so if you feel strongly about it i'll take it out!  ;)

ryanlovett:
@shaneknapp No strong concerns about where to ignore vi swap files. We can ignore them in all the places. :)

shaneknapp:
i have a feeling this file will be a regular WIP ;)

PR #4355, opened 2023-03-10
Based on #4335.

ryanlovett:


ryanlovett:
@d-alex-hughes I was able to merge your changes from #4335 in a separate PR -- CI was having problems with the other one for unknown reasons. Can you please test it at https://ischool-staging.datahub.berkeley.edu and then let us know if it can be merged to production?

d-alex-hughes:
I'm poking around now and things look good! I've got one other student looking from their point of view to see if they can break anything quickly. 

@ryanlovett -- a clarifying question: if we merge this over to `prod` will the students saved work from before the changes persist? There are no files on `ischool-staging` -- but that's reasonable because I haven't done anything on the staging server to this point. 

I just want to double/triple check that I'm not going to wamp 9 weeks of student work.

ryanlovett:
@d-alex-hughes The staging environment has a different set of home directories from the production environment. Nothing is synced between the two. The home storage is also completely orthogonal to the docker images, and merging image changes to prod won't have any impact on student files.

So no wamp worries. The important thing is to test TeX changes and the R package changes that you included in your original PR.

d-alex-hughes:
Those all tested and ran just fine for me and another student. I'm glad to move it to prod.

PR #4362, opened 2023-03-13
Update infer with hot fix

andrewpbray:
Swap version of infer from 1.0.0 on CRAN to a version from github that has a bug fix to rep_sample_n()


ryanlovett:
@andrewpbray Can you test this at https://stat20-staging.datahub.berkeley.edu? If it works the way you intend, I can merge this to production.

PR #4373, opened 2023-03-14
Add docs on specific deployments.

ryanlovett:
The immediate goal was to document stat159, so datahub is essentially a placeholder.


ryanlovett:
This partially addresses #4373.

Issue #4422, opened 2023-04-01
Isolate Data 102 Image from the Main Datahub Image


balajialg:
# Summary

Like #4420 Data 102 instructors ran into a similar problem with Statsmodel and requested a package upgrade in the main hub image. In addition, they requested to isolate the Data 102 image from the Datahub image so that they can make such upgrades whenever necessary. 

# User Stories

- As a `Data 102 instructor` I want to `make changes to Data 102 image` so that `I don't worry about breaking other course workflow`.

# Acceptance criteria

- Data 102 instructors can make changes to their Data 102 hub image which is isolated from the main Datahub image

# Important information

Instructor request 

_Looking at the berkeley-dsep-infra repo, it looks like data102 doesn't have a custom image (i.e., there's no deployments/data102/image analogous to deployments/data100/image). What's the bare minimum necessary to fill that in so that we can have a custom deployments/data102/image/environment.yml file with statsmodels==0.13.5?_

# Tasks to complete

- [ ] Create Data 102 specific image
- [ ]  Reach out to the instructor to verify the urgency of this request (Now or during the maintenance window?)
- [ ] Verify whether Data 102 is taught during summer

balajialg:
- [ ] Reach out to the instructor to verify the urgency of this request (Now or during the maintenance window?)
- [ ] Verify whether Data 102 is taught during summer

balajialg:
Closing this issue as it got done

PR #4423, opened 2023-04-01
Isolate Data 102 Image from Datahub Image & Upgrade Statsmodel Package!

balajialg:
Ref #4422


deployments/data102/hubploy.yaml
```diff
@@ -1,7 +1,7 @@
 images:
   images:
     - name: us-central1-docker.pkg.dev/ucb-datahub-2018/user-images/primary-user-image
```
ryanlovett: The name should be changed too. Most deployment image names are of the form `{deployment}-user-image`. They also go under images.image_name -- see data100 as an example.

deployments/data102/hubploy.yaml
```diff
@@ -1,7 +1,7 @@
...
-      path: ../datahub/images/default
+      path: ../image/
```
ryanlovett: I think if you use images.image_name, you don't need images.path. data100 is also an example. See `git grep image_name` for other deployment examples.

deployments/data102/image/environment.yml
```diff
@@ -0,0 +1,84 @@
...
+- statsmodels=0.13.5
+- scikit-learn=1.1.1
+- seaborn=0.11.*
+- bokeh=2.3.*
+- decorator=5.0.*
+- networkx=2.6.*
+- spacy=3.4.1
+- nltk=3.6.*
+
+# data 88e, fall 2022, issue 3915
```
ryanlovett: I'd remove all dependencies that are associated with courses other than data102. If the course does need one of them later it can be re-added. This helps to keep the image smaller, and reduces the odds you'll have to chase down dependency conflicts should data102 need an update to something.

ryanlovett (changes requested):
The important one is the image name. All other comments are strong suggestions.

balajialg:
Thanks @ryanlovett! Made changes based on your inputs. Let me know what you think?
Should ask the instructor to test their notebooks in staging (When the PR gets merged) to confirm that I did not remove packages necessary for Data 102.

shaneknapp (approved):

ryanlovett:
Yes, once the image is created/deployed it should definitely be tested on staging.

The last question is that this includes desktop apps, but the hub isn't using the jupyter desktop proxy. If the class doesn't need the remote desktop functionality, then the following should be removed:

```
## This section adds tools for desktop environment usage
dbus-x11
xorg
xubuntu-icon-theme
xfce4
xfce4-goodies
xclip
xsel
firefox
chromium-browser

# GUI text editors
emacs
vim-gtk3
gedit
```

balajialg:
Closing this PR sadly as there is no response from the instructor @rameshvs despite repeated outreach!

cc: @shaneknapp

PR #4434, opened 2023-04-04
fail when we are unable to get the commit range in circleci

shaneknapp:
currently, if we are unable to determine the commit range that we're testing against, the script will exit w/a retval of 0 and continue the build and fail later on.

i propose failing if there's no commit range.  :)

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,6 +46,7 @@ orbs:
                         if [ -s /tmp/commit-range-err.txt ]; then
                             echo "Unable to get commit range."
                             cat /tmp/commit-range-err.txt
+                            exit 1
                         fi
                         echo COMMIT_RANGE: ${COMMIT_RANGE}
                         echo "export COMMIT_RANGE='${COMMIT_RANGE}'" >> ${BASH_ENV}


shaneknapp:
```
github.GithubException.RateLimitExceededException: 403 {"message": "API rate limit exceeded for 54.224.184.90. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)", "documentation_url": "https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
```

maybe we could authenticate?

ryanlovett:
@shaneknapp Later in the file, hubploy will try to workaround the case where COMMIT_RANGE is empty, in `Build image if needed`.

One way of dealing with this is to catch the exception in get-commit-range.py, rather than using `|| true` which apparently isn't enough to trap CI errors. Then `hubploy` may do the right thing later on.

But ultimately I think we need to auth to github to avoid the rate limits. Long term maybe we should move to Github Actions.

shaneknapp:
> 

word, i'll open a new PR and figure this one out.

PR #4456, opened 2023-04-10
Update to JupyterLab 3.6.3

fperez:
This should fix some of the document corruption issues we've experienced recently.


fperez:
Thanks @ryanlovett!

Issue #4472, opened 2023-04-12
Request more RAM for class data 200


rohanjha123:
**Which hub do you want more RAM on? Please link the URL to the hub while specifying the hub name**

Data 100 Hub (data100.hub.berkeley.edu)

**Which class is this request for?**

Data 200, spring 2023

**What is the Academic Guide URL for this class?**

Data C200: https://classes.berkeley.edu/content/2023-spring-data-c200-001-lec-001 
Data 200S: https://classes.berkeley.edu/content/2023-spring-data-200s-001-lec-001

**What is the bCourses URL for this class?**

https://bcourses.berkeley.edu/courses/1521765

**How many students do you expect in this class?**

90

**How much RAM per user is needed?**

8GB ram per student

**Why does this class need this much RAM?**

The graduate students are working on a grad project, which requires extra RAM

**By what date (MM/DD) do you want this request to be fulfilled?**

ASAP if possible, graduate students have already begun working on the project! Sorry for the late request

**Any additional information we should know about?**

There's no need to increase the RAM for students in data C100; only grad students need extra RAM

ericvd-ucb:
8gb. What assignment is this for?

rohanjha123:
The graduate student project! Previously, the grad students had to figure out google collab/git to work with the larger datasets, but now we want to also let them work on datahub if they wish

balajialg:
@rohanjha123 Do you have a deadline to revert this memory increase? (As you may know already, this request correlates with increased cloud costs. Hence, want to have a strict deadline to revert such changes)

yanlisa:
Hi @balajialg, here is our graduate project timeline: https://ds100.org/sp23/gradproject/#timeline I'm thinking that if students are currently all at 2GB, we could increase to 4GB starting 4/17 and then increase again to 8GB 4/21-5/14 (+2 days after final writeup to accommodate late penalties). Would this be an overwhelming increase to cloud costs? Happy to discuss more. Thanks

balajialg:
@yanlisa Thanks for the detailed timeline! I think this should work. We will bump up the resources accordingly. We can have a short conversation about cloud costs during next week's meeting.

PR #4474, opened 2023-04-13
Increase memory requirement to 4 GB for Data 200 students 

balajialg:


balajialg:
Fixes #4472

ryanlovett (commented):
That is the correct syntax. Does it need an altered ramp up and end date like in #4472?

balajialg:
@ryanlovett yes - I will update the config and add a new commit in this PR soon

balajialg:
@shaneknapp Increased the RAM to 4GB as per the instructors request [here](https://github.com/berkeley-dsep-infra/datahub/issues/4472#issuecomment-1506162111). Can you merge this to staging?

shaneknapp (approved):

PR #4511, opened 2023-04-24
Documentation updates informed by recent work

gmerritt:
Updates to new-hub, rebuild-hub, and preview.

### DIFF
--- a/docs/admins/howto/new-hub.rst
+++ b/docs/admins/howto/new-hub.rst
@@ -171,11 +171,7 @@ Check whether the directories have permissions similar to the below directories:
 
 Create the hub deployment locally
 ---------------------------------
-In the ``datahub/deployments`` directory, run ``cookiecutter``. This sets up the hub's configuration directory.
-You may do this interactively, responding to prompts, or you may do this non-interactively by providing the settings
-in a hub-specific configuration file.
-
-**Interactively:**
+In the ``datahub/deployments`` directory, run ``cookiecutter``. This sets up the hub's configuration directory:
 
 .. code:: bash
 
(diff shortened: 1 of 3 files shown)


docs/admins/howto/preview-local.rst
```diff
@@ -23,4 +23,17 @@ Steps to preview locally as HTML files
...
-#. Have fun making changes to the documentation based on the HTML preview.
\ No newline at end of file
+#. Have fun making changes to the documentation based on the HTML preview.
+
+Strategy 2: free online previewer
+======================================
+
+Use a free, online, in-line converter such as: https://www.tutorialspoint.com/online_restructure_editor.php
+
+Strategy 3: Visual Studio code extension
```
shaneknapp: another strategy that i'd strongly recommend including is:
```
$ conda install sphinx

$ rst2html.py <filename>.rst > <filename>.html
```

docs/admins/howto/new-hub.rst
```diff
@@ -188,6 +192,27 @@ The cookiecutter template will prompt you to provide the following information:
 
 This will generate a directory with the name of the hub you provided with a skeleton configuration and all the necessary secrets.
 
+**Non-interactively:**
```
ryanlovett: Editing a template to run non-interactively and then restoring it doesn't feel like the right approach, because one may accidentally forget to restore the template. I think it'd be better to seed the additional information without modifying the template. Based on [cookiecutter docs](https://cookiecutter.readthedocs.io/en/stable/advanced/suppressing_prompts.html), it looks like one can provide values non-interactively through a wrapper script.

Untested example:
```python
import json
from cookiecutter.main import cookiecutter
deployment_values = json.loads(open('newhub/config/cookiecutter.json').read())
cookiecutter('template/', no_input=True, extra_context=deployment_values)
```

And this would be run in `deployments/` as `../scripts/hubcutter.py` or some such.

ryanlovett (commented):
Thanks for adding this material! I had one thought about the cookiecutter edit/restore process and suggested an alternative that would require a (seemingly) small amount of coding.

docs/admins/howto/new-hub.rst
gmerritt: Thanks, @ryanlovett!

Yeah what's captured in that document (including the "template musical chairs") is definitely more of a "what we did that day" rather than an intentional design.

I'll note that there is a notion to automate/script much or all of new hub deployment at some level, and both the cookiecutter "template musical chairs" and the more-direct proposed hubcutter.py can be considered as bridging the gap between the prompt/response approach and the possible end goal of the fully-scripted hub creation.

Given the (possible) trajectory toward a more-unified & scripted new hub deployment process, the cookiecutter changes being discussed here might only be a transient/temporary recommended workflow. Should we perhaps leave the cookiecutter documentation as-is? I.e., if any of us were to make a new hub tomorrow (long before a more-fully-automated process is available), how would we choose to do it? The old prompt way, or one of the other ways?

ryanlovett: It sounds like there's not a lot of time pressure since there are no new hubs imminent. I think the script might take 20m-60m which, assuming I've read the cookiecutter docs right, seems like it'd be worthwhile and fairly quick. If I did need to make a new hub tomorrow I'd try writing it.

Otherwise I'd amend the docs to say that there's some danger that the template could be accidentally altered.

gmerritt:
```python
deployment_values = json.loads(open('newhub/config/cookiecutter.json').read())
```

One thing I don't get: Isn't the job of `cookiecutter` to create the `newhub/` directory and its skeleton contents? If so, how would `hubcutter.py` read from `newhub/config/cookiecutter.json` if `newhub/config/` does not yet exist?

Or maybe I'm missing something, for sure!

[Edit: misplaced backtick repositioned]

ryanlovett:
> One thing I don't get: Isn't the job of cookiecutter to create the newhub/ directory and its skeleton contents? If so, how would hubcutter.py read from newhub/config/cookiecutter.json if newhub/config/ does not yet exist?

Yeah, good point. It'd need to be somewhere else initially, or the params could be provided as CLI args.

shaneknapp (changes requested):

shaneknapp (approved):

PR #4515, opened 2023-04-26
magically add documentation label to new prs

shaneknapp:


gmerritt:
I totally dig it, and am down with the change -- although I cannot double-check the implementation/code, as I am not familiar with using labeler.yml etc.

shaneknapp:
not sure if we need a github auth token, but i think we'll find out once this is merged to staging.

shaneknapp:
> I totally dig it, and am down with the change -- although I cannot double-check the implementation/code, as I am not familiar with using labeler.yml etc.

the globs are the same regex format as used in `.gitignore` etc...  so that's pretty straightforward.

but this is also my first time adding a github thingy to the thingy, and i cargo-culted `workflows/labeler.yml` from https://github.com/marketplace/actions/labeler

PR #4531, opened 2023-04-27
[DH-21] adding more labels for code, config and image changes

shaneknapp:
i added labels for code (R, python) and "configuration" and images.

i'd really like some feedback on this at some point before merging in...  it will be nice to have some labels but i don't necessarily want to overdo things.  :)

this is also not to be considered "high priority"...

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,3 +1,31 @@
+# add labels for code changes
+python:
+  - "**/*.py"
+R-code:
+  - "**/*.R"
+bash:
+  - "**/*.sh"
+  - "**/*.bash"
+  - "**/postBuild"
+
+# add labels for config changes
+configuration:
+  - "**/*.yml"
+  - "**/*.yaml"
+  - "**/*.json"
+  - "**/*.r"
+  - "**/Dockerfile"
+  - "**/*.conf"
+  - "**/requirements.in"
+  - "**/requirements.txt"
+  - "**/infra-requirements.txt"
+  - "**/*.site"
+  - "**/*.jl"
+
+# images
+images:
+  - "**/*.png"
+
 # add labels to docs
 documentation:
   - "docs/**"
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
```diff
@@ -1,3 +1,31 @@
...
+  - "**/*.bash"
+  - "**/postBuild"
+
+# add labels for config changes
+configuration:
+  - "**/*.yml"
+  - "**/*.yaml"
+  - "**/*.json"
+  - "**/*.r"
+  - "**/Dockerfile"
```
shaneknapp: re Dockerfile...  wondering if this should have a 'code'-based label, config, or both?

ryanlovett: My tendency would be to do what you're doing here and try to tag like things. That being said, I'm trying to imagine cases where I'd need to go to our issues/PR pages and filter by some of these. Perhaps its more for assessment and reporting purposes rather than development or support. I think it'd be worthwhile to add these since it does not require a lot of resources for labeler to run, and it doesn't take up head space to manage all of them.

In the above, `**/*.r` is actually R code and not configuration.

I think that `Dockerfile` should be `code` or `dockerfile` or `image`, but not `configuration`. The latter applies to files that are for hub configuration. Maybe it should be subsumed by a `**/image` match for an `image` label.

ryanlovett: Or maybe `singleuser` instead of `image` to avoid confusion with `images` below.

shaneknapp: > Or maybe `singleuser` instead of `image` to avoid confusion with `images` below.

yeah.  or `singleuser-image`...

i can do some cool stuff like:
```
singleuser-image:
  - "deployments/**/image/**"
  - "deployments/**/images/**"
```
this will catch all single user image configs etc, including png/svg included in said image configs.

now i can explicitly ignore these same paths and add the `configuration` label on all other `y(a)ml` files.  this means that `configuration` would be tied to non-user image configs in datahub:
```
configuration:
  - "!deployments/**/image/**"
  - "!deployments/**/images/**"
  - "**/*.yml"
  - "**/*.yaml"
  - "**/*.json"
```

PR #4532, opened 2023-04-27
[DH-27] first pass at fixing dependabot alerts

shaneknapp:
more info here:  https://github.com/berkeley-dsep-infra/datahub/security/dependabot

updated everything to the min recommended version...

### DIFF
--- a/deployments/datahub/images/default/requirements.txt
+++ b/deployments/datahub/images/default/requirements.txt
@@ -58,7 +58,7 @@ dustmaps==1.0.9
 george==0.4.0
 exoplanet==0.5.2
 torch==1.13.1
-torchvision==0.13.1
+torchvision==0.14.0
 pyvo==1.1
 joblib==1.0.1
 theano-pymc==1.1.2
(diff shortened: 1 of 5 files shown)


images/node-placeholder-scaler/requirements.txt
shaneknapp, line 1: i tested this locally w/python3.9 and the calendar scaler still works.

deployments/datahub/images/default/requirements.txt
```diff
@@ -57,8 +57,8 @@ astropy==5.1
 dustmaps==1.0.9
 george==0.4.0
 exoplanet==0.5.2
-torch==1.12.1
-torchvision==0.13.1
+torch==1.13.1
```
shaneknapp, line 60: hey @balajialg is this course still using datahub?

`astr 128/256; spring 2021`

balajialg: @shaneknapp Seems like this course is not currently offered this semester - https://classes.berkeley.edu/search/class/astronomy%20128?retain-filters=1&f%5B0%5D=im_field_term_name%3A2729

PR #4573, opened 2023-05-15
Remove config changes across hubs related to RAM increase!

balajialg:


ryanlovett:
@balajialg Looks like the datahub and eecs changes were done by #4572 in https://github.com/berkeley-dsep-infra/datahub/pull/4572/commits/e8ea46f270c3e006aa1c4d6740a6dc3d9f6d6d19. The data100 change hasn't been done yet. (was an oversight on my part)

balajialg:
@ryanlovett Ah, I didn't know you removed memory increase related changes as the commit name was titled "remove admin names". No worries and thanks for doing that, I will update this commit with just Data 100 hub-related config changes.

ryanlovett:
@balajialg Sorry, I should have use a separate commit to emphasize what action was being performed, or mentioned it in the description.

balajialg:
@ryanlovett Yes - adding some info about the commit in the description would be super helpful. Thanks again!

PR #4578, opened 2023-05-16
Remove python-mode from a11y.

ryanlovett:
It isn't present in jammy and isn't needed for a11y. (nor is gfortran)


ryanlovett:
Regarding jammy, I'm testing a new version of repo2docker which supports jammy. We need to determine which bionic packages are no longer present in later versions.

Issue #1505, opened 2020-06-08
Request python package PyQt5, PIL for class EECS 16A


Quizenger:
**Which package do you want installed?**
PyQt5 to display visual demonstrations using a GUI/Graphics window. This opens a separate window outside the jupyter notebook interface and we need it to be visible (I'm not sure if that's possible with DataHub)
PIL (Pillow fork) to open, save and modify images

**Which hub do you want it installed on?  For example r.datahub, datahub, data100.datahub, etc.**
eecs.datahub

**Which class will use this package?**
EECS 16A

Include a link to appropriate entry in https://classes.berkeley.edu/ if available. If not, please mention class name & instructor.
https://classes.berkeley.edu/content/2020-summer-eecs-16a-001-lec-001
https://classes.berkeley.edu/content/2020-fall-eecs-16a-001-lec-001

**Which semester will this package be used for?**
Summer 2020 (urgent), Fall 2020

This helps us clean up unused packages after a term ends.

**Any additional information we should know about?**

yuvipanda:
Hello @Quizenger!

I've added pillow to eecs.datahub.berkeley.edu. PyQT5 doesn't work via a browser generally - however, we added a special 'desktop' view that you can try out at the eecs datahub. This gives you a full linux desktop inside the browser, although we haven't given it a lot of RAM yet. If you can tell me what PyQT5 was used for, we can try to see if it works there?

Quizenger:
Hi @yuvipanda!

Thanks for adding Pillow! The desktop view idea sounds great! 

I tried testing it out (by generating an nbgitpuller link for my public repo) but it showed the PyQt5 module as missing and I couldn't find the desktop view. Am I supposed to try the changes at eecs datahub some other way?

PyQt5 is used to create a few QtWidgets that finally lead to one window being launched that displays lots of matrices (visualizing matrix computations) in a timer-fired loop.
Feel free to have a look at cumulative_imaging_simulation.py using either of the following links!

nbgitpuller link: https://eecs.datahub.berkeley.edu/hub/user-redirect/git-pull?repo=https%3A%2F%2Fgithub.com%2FQuizenger%2Ftesting_datahub&urlpath=tree%2Ftesting_datahub%2Fee16a_imaging_lab3%2Fee16a_imaging_lab3_nOMP

My public repo link: https://github.com/Quizenger/testing_datahub/tree/master/ee16a_imaging_lab3/ee16a_imaging_lab3_nOMP

[...]

PR #4587, opened 2023-05-18
upgrading data100, eecs, stat159 to jammy

felder:


felder:
@Quizenger A number of the hub images are using ubuntu bionic which reaches end of life at end of this month

Unfortunately the eecs image installs qt5-default which is a package no longer available in the latest ubuntu LTS:  
https://stackoverflow.com/questions/74110674/problem-installing-qt5-default-on-ubuntu-22-04

Is this package still a dependency for your course?  It was added as part of work done in:
https://github.com/berkeley-dsep-infra/datahub/issues/1505

I'm going to install the recommended replacements, but you'll probably want to test whatever functionality this package provides.  If you no longer require it, I'll just remove it.

PR #4592, opened 2023-05-24
fix labels for docs

shaneknapp:
looks like we needed to put all of the `.txt` conditions in one `any` statement


shaneknapp:
tested by https://github.com/berkeley-dsep-infra/datahub/pull/4593

PR #4614, opened 2023-06-05
add section for expected turnaround

shaneknapp:
@balajialg does 5-10 business days seem reasonable?


balajialg:
@shaneknapp Yes - I am aligned. Thanks for initiating this! We should keep 10 business days as hard limit given the list of things to be figured out when Caesar transitions off this project.

balajialg (approved):

Issue #4613, opened 2023-06-05
[Data 8 - Summer 2023] Admin Request


Kevin-Miao:
### Your Name

Kevin Miao

### Affiliated Course Name

Data 8

### Role

Instructor

### Name of the Hub

data8.datahub.berkeley.edu

### List of Usernames

[ moved by @ryanlovett to Google Drive ]

### End Date

08/15/2023

balajialg:
Thanks, @Kevin-Miao for the request! Going forward, we are following a new process to assign admins for courses. We will enable course instructors and GSIs affiliated with Data 8 (as listed in bcourses) to be admins directly. Can you please let us know if anyone who is not not affiliated with the course in bcourses as a teacher/ta should be given admin access?

**Reference:** https://docs.datahub.berkeley.edu/en/latest/admins/howto/course-config.html#defining-group-profiles
cc: @ryanlovett

Kevin-Miao:
@balajialg Awesome thanks for the fyi. Question: is this sync direct and are any additional requirements necessary to make this work?

balajialg:
@Kevin-Miao I think the information we need from you are a) bcourses id for Data 8 and b) anyone who is not not affiliated with the Data8 in bcourses as a teacher/ta but should be given admin access. With this info, We can easily enable admin access directly through the config files.

Kevin-Miao:
Hi Balaji,

The bCourses ID should be the SYS ID right? For the summer data 8 term, this is `CRS:DATA-C8-2023-C`. The ID in the url corresponds to `1525580`. All staff members with the roles teacher, TA, lead TA, and Reader should be able to have access. We are not expecting to have any external admins requiring access at this point. Thanks!

balajialg:
@Kevin-Miao Created a PR to enable admin access for Data 8 instructional staff. Will give you a heads up once it gets merged to prod. Thanks for your info!

balajialg:
Closing this issue as it got completed through #4637

PR #4615, opened 2023-06-05
Modify github template for admin request to ask for bcourses id

balajialg:
Based on this request: #4613


ryanlovett (changes requested):
The first part looks good to me, although it'd be better if the username list could be communicated through the ds-infrastructure mailing list (IMO).

The data8/config/common.yaml change should not be a part of this PR since it is unrelated to the template change. Also`jupyterhub.singleuser.admin.group_profiles.admin=True` does not do anything.

balajialg:
@ryanlovett Made this PR only about changes to the admin access template. Incorporated your feedback about requesting usernames via ds-infra list. ]

Will create a new PR for enabling admin access in Data 8 hub

PR #4632, opened 2023-06-09
[DH-54] upgrade python on calendar autoscaler to 3.11

shaneknapp:
https://jira-secure.berkeley.edu/browse/DH-54

tested everything and it works as expected...


ryanlovett (approved):
Looks straightforward. +1 on testing under 3.11 too.

PR #4637, opened 2023-06-12
Enable admin for Data 8.

ryanlovett:


shaneknapp:
i'm curious -- does the `custom:` block need to live at the very end of `common.yaml`?

ryanlovett:
@shaneknapp No, anything reading the yaml should just see it as a key under `hub`.

PR #4657, opened 2023-06-15
removed older pinned versions of repo2docker, pin to latest release

shaneknapp:
@ryanlovett am i missing anything else?

### DIFF
--- a/deployments/data100-jl4/image/environment.yml
+++ b/deployments/data100-jl4/image/environment.yml
@@ -30,7 +30,7 @@ dependencies:
 - ipyparallel==8.4.1
 - jsonschema==4.17.3
 - jupyter-book==0.15.1
-- jupyter-repo2docker==2022.2.0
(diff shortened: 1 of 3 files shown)


ryanlovett:
@shaneknapp I don't think repo2docker is actually needed inside the deployment images, just in requirements.txt. As far as I know, nobody is building images from within a hub. So my sense is rather than upgrade r2d within data100/data100-jl4, you can just remove it.

shaneknapp:
+1 sgtm!

ryanlovett (approved):
Looks good! Give 2023.06.0 a try locally if you haven't already.

We can then release a new hubploy and bump that.

shaneknapp:
> Looks good! Give 2023.06.0 a try locally if you haven't already.
> 
yep!  i built data100-jl4 in like 30sec and it worked like a charm.

> We can then release a new hubploy and bump that.

awesome

PR #4665, opened 2023-06-21
Removing jupyter themes package with disappointment

balajialg:
Realized during exploration with @ryanlovett that a) this tool changes the default theme and b) There is no option to revert to the old theme. Removing this package for this reason.

In addition, I am exploring whether the URL issue is caused by JN 7.0rc0? I want to eliminate unnecessary packages to make a claim


balajialg:
Thanks @ryanlovett

PR #4688, opened 2023-06-27
Add temporary development R hub

ryanlovett:
I'm creating this hub to explore a [jupyter-rsession-proxy issue](https://github.com/jupyterhub/jupyter-rsession-proxy/issues/140).

While I was at it, I thought I'd explore multi-stage docker builds, eyeing the main datahub image as a potential use-case. This explains why this image looks like that one, albeit with many fewer dependencies. I also steered it towards repo2docker by moving dependencies from Dockerfile into apt.txt and environment.yml.

hubploy uses docker-py which does not support Docker BuildKit, so we can't take advantage of the multi-stage build concurrency yet. I may look at adding subprocess execution to hubploy to enable this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -317,12 +317,18 @@ jobs:
           command: |
             hubploy deploy --timeout 30m eecs hub ${CIRCLE_BRANCH}
           no_output_timeout: 30m
-          
+
       - run:
           name: Deploy a11y
           command: |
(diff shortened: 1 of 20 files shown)


shaneknapp:
@ryanlovett you should also add this to the github labeler...  https://docs.datahub.berkeley.edu/en/latest/admins/howto/new-hub.html#add-hub-to-the-github-labeler-workflow

PR #4729, opened 2023-07-10
Fix pandoc version

shaneknapp:


shaneknapp:
i hate this crap.  pip only allows up to pandoc 2.3, the env file initially asks for 2.19.2 which doesn't seem to be available.  `conda search` shows that 2.12 is out there, but the official conda docs show that [3.1.3 should be availble](https://anaconda.org/conda-forge/pandoc) (but isn't).  :headdesk:

PR #4733, opened 2023-07-11
Give the 140 TAs admin access again

shahzarrizvi:
140 TAs lost admin access; add the TAs for Fall 2023 back.


ryanlovett:
Hi @shahzarrizvi , what is Fall '23 Prob 140's bcourse course ID? (the number in the URL)

ryanlovett:
Closed in favor of #4740. We can revisit if there are issues with role loading.

PR #4763, opened 2023-07-17
strip html from event description

shaneknapp:


shaneknapp:
tested this locally w/the correct python environment etc.  works like a charm!

images/node-placeholder-scaler/scaler/scaler.py
```diff
@@ -137,6 +137,8 @@ def main():
                 ):
```
shaneknapp, line 137: see the line above (line 136):  i just double checked the logs and this conditional is NEVER met!  the actual output from the `kubectl` execution always returns `Setting XXXX to Y replicas`, even if the command doesn't change anything.
Screenshot 2023-07-18 at 2 16 22 PM

shaneknapp:
well, i chartpressed this and it still doesn't work.  from the logs today:
image
image

at the very least i now have a calendar entry to specifically test against!  :)

shaneknapp:
following my own instructions is hard.  merging, and i will confirm once it's deployed fully.

output of my testing from the interpreter:
```
Python 3.11.4 (main, Jul  5 2023, 13:45:01) [GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import scaler.calendar
>>> import datetime
>>> import zoneinfo
>>> tz = zoneinfo.ZoneInfo(key='America/Los_Angeles')
>>> html_event = datetime.datetime(2023, 7, 19, 9, 0, 0, tzinfo=tz)
>>> calendar = scaler.calendar.get_calendar('https://calendar.google.com/calendar/ical/c_s47m3m1nuj3s81187k3b2b5s5o%40group.calendar.google.com/public/basic.ics')
>>> events = scaler.calendar.get_events(calendar, time=html_event)
ev.descrition before removing html:<span>datahub-pool</span>: 2
>>> events[0].description # after html is stripped
'datahub-pool: 2'
```

shaneknapp:
ok it worked for me:  
image

ill have balaji create an event and confirm there as well.

PR #4766, opened 2023-07-19
remove grafana-spamming logic bomb from the calendar autoscaler

shaneknapp:
yep, never trust STDOUT and negative logic.  we have been spamming grafana with `<number of hubs> * 60` annotations probably since for-ev-er.

right now, let's disable this.  maybe this was impacting grafana?  @ryanlovett and i poked around a few months back but never found anything there...  some (more robust) logic around things changing and annotations might be useful later down the road.

:headdesk:

### DIFF
--- a/images/node-placeholder-scaler/scaler/scaler.py
+++ b/images/node-placeholder-scaler/scaler/scaler.py
@@ -139,12 +139,13 @@ def main():
                 # with a notation for each hub, once a minute, for perpetuity.
                 #
                 # the 'Find out what happened her' is, i assume, a breadcrumb
-                # yuvi, also leaving me to believe that this never really
-                # worked as intended.
+                # from yuvi, also leaving me to believe that this never really
+                # worked as intended. ;)
                 #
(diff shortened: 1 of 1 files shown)


images/node-placeholder-scaler/scaler/scaler.py
```diff
@@ -131,14 +131,21 @@ def main():
...
 
                 logging.info(proc.stdout.strip())
-                if (
-                    proc.stdout.strip()
-                    != f"deployment.apps/{pool_name}-placeholder unchanged"
-                ):
-                    # Something has changed, let's signal that
-                    actions_taken.append(f"{pool_name} set to {replica_count}")
-                    f.seek(0)
-                    logging.info(f"yaml passed to kubectl: {f.read()}")
```
shaneknapp, line 141: fwiw, me dumping the yaml to the logs isn't very useful...

shaneknapp:
actually -- according to `git blame`, it's been spamming since sept 2, 2021 XD

PR #4794, opened 2023-07-27
[data8x] Removed

sean-morris:
I commented out the changes because after it is removed I am going to comment back in data8x but it will be replaced by what is in data8xv2 but renamed to edx


shaneknapp:
fyi the data8xv2 builds were failing and i needed to comment them out in https://github.com/berkeley-dsep-infra/datahub/pull/4799

felder:
See #4922 as that was easier than trying to deal with the conflicts here.

PR #4808, opened 2023-08-01
reactivating cee and astro as per DH-64

felder:
https://jira-secure.berkeley.edu/browse/DH-64

Looks to me like data102 is already deployed.


balajialg:
@felder As updated in DH-64, it is Data 101 hub that needs to get recommissioned!

felder:
Let's go ahead and merge this (assuming it looks right) and then I'll submit another PR for data101.

shaneknapp (approved):

PR #4811, opened 2023-08-02
Disable stat159 in CI.

ryanlovett:
It won't be taught for at least a year.


ryanlovett:
Not sure why upstreams are unable to find this branch, but it should be safe to merge.

PR #4819, opened 2023-08-03
remove cesar from the archive retrieval request issue template, add balaji

shaneknapp:
@balajialg adding you to this template since you'll be helping out this month.

not sure if i should remove felder and myself yet, especially since the new tool makes this task relatively straightforward and easy to do (and therefore easy for us to jump in and help if needed).


balajialg:
Thanks! I think you can remove @felder from this template as he clearly expressed that he is not interested to handle these requests. You can still be there in this template so that you can handle some of these requests when the volume is higher than expected this month.

PR #4820, opened 2023-08-03
add labels for  LICENSE, github issue template, helm/chart , and support and hub deployment changes

shaneknapp:
i missed these two dirs in our repos and they should most certainly have their own labels!

@ryanlovett @felder any thoughts on the naming of these labels?

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -44,7 +44,10 @@ documentation:
 # add build-infra label to any .github or circleci changes
 build-infra:
   - ".circleci/**"
-  - ".github/**"
+  - any: ['.github/**', '!.github/ISSUE_TEMPLATE/**']
+
+issue-templates:
+  - ".github/ISSUE_TEMPLATE/**"
 
 # support systems
 support-deployment:
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -46,6 +46,14 @@ build-infra:
   - ".circleci/**"
   - ".github/**"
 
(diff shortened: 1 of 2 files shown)


shaneknapp:
once we agree on label names, i'll use this PR as an opportunity to normalize to single quotes instead of a mix of those and double.

ryanlovett:
Do you want to label all chart-related matters? e.g. chartpress.yaml, Chart.yml, values.yaml, etc. This would intersect with some of the others but could be useful for search.

`documentation` doesn't catch README.md or LICENSE at the top level, although they're probably not going to change.

shaneknapp:
> Do you want to label all chart-related matters? e.g. chartpress.yaml, Chart.yml, values.yaml, etc. This would intersect with some of the others but could be useful for search.
> 
yes, great idea.

> `documentation` doesn't catch README.md or LICENSE at the top level, although they're probably not going to change.

actually it does!  `**/*.md` globs for any `md` files at any level of the repo hierarchy.  :)

ryanlovett:
> actually it does! **/*.md globs for any md files at any level of the repo hierarchy. :)

Ahh, TIL! When then just LICENSE, but that's not critical.

shaneknapp:
> > actually it does! **/*.md globs for any md files at any level of the repo hierarchy. :)
> 
> Ahh, TIL! When then just LICENSE, but that's not critical.

not critical, but quite easy...  :)

Issue #4822, opened 2023-08-03
Request more RAM for class ESPM-157


cboettig:
**Which hub do you want more RAM on? Please link the URL to the hub while specifying the hub name**

R Hub (r.datahub.berkeley.edu hub)

**Which class is this request for?**

For example, "2023 Fall: ESPM-157: Reproducible and Collaborative Data Sciences".

**What is the Academic Guide URL for this class?**

https://classes.berkeley.edu/content/2023-fall-espm-157-001-lab-001

**What is the bCourses URL for this class?**

https://bcourses.berkeley.edu/courses/1527100

**How many students do you expect in this class?**

54

**How much RAM per user is needed?**

4 GB ram

**Why does this class need this much RAM?**

This course involves four modules that each seek to reproduce key papers in global change using the most recent available data and open source software.  While we make deliberate use of out-of-core methods that allow us to minimize RAM use (arrow, duckdb, gdal), some steps (especially visualization) still need about 4 gb to reliably execute without crashing the environment.  

In a pinch we can try to have students use their laptops, since most machines these days have 4 gb free RAM, but historically the university wifi has not proven very suitable to having 54 students do streaming data processing....  the hub has been so much better for this! (not to mention sidestepping installation issues)  

**By what date (MM/DD) do you want this request to be fulfilled?**

09/01

**Any additional information we should know about?**

Memory increased required till the finals

balajialg:
Closing this issue as it got resolved. Please do reopen if that is not the case

PR #4823, opened 2023-08-03
Increase RAM for ESPM 157

balajialg:
https://github.com/berkeley-dsep-infra/datahub/issues/4822

### DIFF
--- a/deployments/r/config/common.yaml
+++ b/deployments/r/config/common.yaml
@@ -71,3 +71,9 @@ jupyterhub:
       guarantee: 512M
       limit: 1G
     image: {}
+
+    custom:
+      group_profiles:
+        1527100: # ESPM 157, Fall 2023, issue #4822
+          mem_limit: 4096M
+          mem_guarantee: 4096M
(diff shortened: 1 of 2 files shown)


deployments/r/config/common.yaml
```diff
@@ -71,3 +71,9 @@ jupyterhub:
       guarantee: 512M
       limit: 1G
     image: {}
+
+    custom:
+      group_profiles:
+        1527100: # ESPM 157, Fall 2023, issue #4822
```
shaneknapp: i think this needs to be `course::1537100`.  @ryanlovett to confirm his docs are correct...  :)

https://docs.datahub.berkeley.edu/en/latest/admins/howto/course-config.html#defining-group-profiles

shaneknapp (changes requested):

balajialg:
@shaneknapp Makes sense, just appended "course::" in the syntax! Thanks

deployments/r/config/common.yaml
ryanlovett: Yep! Using the course ID alone was a previous method.

Issue #4843, opened 2023-08-10
Data 101 Hub Setup Requirements for FA 23!


balajialg:
# Summary
@yanlisa will be teaching [Data 101 course](https://classes.berkeley.edu/content/2023-fall-data-101-001-lec-001) using [Data 101](https://data101.datahub.berkeley.ed) hub during FA 23. 250 students will be using this hub during the duration of the semester. Data 101 will have [4 projects ](https://fa22.data101.org/) during the entire semester and all of them will involve dealing with databases (the first 3 assignments will use PostgreSQL database and the last assignment will use Mongo database). [Here](https://data101.datahub.berkeley.edu/user/balajialwar/notebooks/fa22-materials/proj/proj1/proj1.ipynb) is an example assignment from this course.

Listed below are the requirements for this course,

- [x] Students should be able to launch Postgres and Mongo db instances (Currently this is commented out in the [configs](https://github.com/berkeley-dsep-infra/datahub/blob/staging/deployments/data101/config/common.yaml))

- [x] Create Data 101 image (Currently, it is using Datahub's [image](https://github.com/berkeley-dsep-infra/datahub/blob/staging/deployments/data101/hubploy.yaml))

- [x] Make JupyterLab4 the default interface for Data 101 hub

- [x] Ensure that JupySql package is installed in the image

- [x] Increase the default RAM requirement to 2 GB

Course staff is planning to test the hub early next week. In addition, There will be a meeting scheduled with the GSIs to onboard them (Similar to the training with Data 100 GSIs)

Students access the data in this hub in the following way, 
a) Students will download the zipped version of the required data from Github repositories using nbgitpuller, 
b) They will unzip the file and store the data in a PostgreSQL database and 
c) Use Jupyter notebook to retrieve the data from database. 

@yanlisa believes that this process is inefficient as multiple copies of the same file exist in each user's hub. She is hoping that a long-term requirement will be to explore the idea of shared database instances (Mongo and PostgreSQL) where students can interact with the data directly (w/o downloading it to their home directories). This is something we need to explore during our internal meeting and have a discussion with Lisa during mid semester.

shaneknapp:
edited coz i was wrong:  data101 will need it's own image, or data100's image will need to have mongodb and postgress installed.  any db config will need to be custom for data101.

i'd strongly suggest having a unique data101 image.  this is truly not a lot of work, but the devil's in the details -- especially if setting up the dbs doesn't go smoothly.

[...]

PR #4850, opened 2023-08-11
uncommenting db configs for DH-94

felder:


shaneknapp:
https://github.com/berkeley-dsep-infra/datahub/issues/4843

PR #4852, opened 2023-08-11
Basing image off data100-jl4 as per DH-93

felder:


shaneknapp:
https://github.com/berkeley-dsep-infra/datahub/issues/4843

PR #4904, opened 2023-08-21
Update quarto, sync libs from textbook image.

ryanlovett:


andrewpbray:
Is the plan to have the image for stat20.datahub pull from the stat20 curriculum image at dockerhub? and will it have it's own .r file to load packages or will it just use whatever is there on the existing image? Or are we just going to try to manually keep the two images similar through the Quarto version and the r packages?

One thing this makes me think of: we should probably peg the rocker verse image to a specific tag at this point instead of just pulling latest, right?

ryanlovett:
@andrewpbray Right now the plan is to just have the datahub image support what you have been doing with the textbook. Per last week's community meeting, we want to be able to support images external to this repo, but we're not there yet.

I think both images are based on 4.3.1 ([datahub](https://github.com/berkeley-dsep-infra/datahub/blob/staging/deployments/stat20/image/Dockerfile#L1), [stat20-docker](https://github.com/stat20/stat20-docker/blob/main/Dockerfile#L4)), but the [workflows](https://github.com/stat20/stat20/blob/main/.github/workflows/publish-staff-site-pr.yml#L19) for stat20 use `latest`, so they should be pegged.

One strategy to support ease of development on your side while retaining the ability to pull in infrastructure updates on our side could be around [multi-stage builds](https://docs.docker.com/build/building/multi-stage/). I experimented with this in the dev-r deployment in this repo. The idea would be that both images share a common base, and the datahub image pulls in assets from the external one.

Issue #4913, opened 2023-08-22
Data 101 Admin Access Request


frank-lsf:
### Your Name

Frank Luan

### Affiliated Course Name

Data101

### Bcourses Id

1528038

### Role

GSI

### Name of the Hub

Data101

### End Date

- 
12/31/2023

balajialg:
@franklsf95 Closing this request as the changes got merged to prod. Please reopen if you still don't have any admin access

PR #4915, opened 2023-08-22
Add admin users for Data 101 Hub

balajialg:
First try at this request - https://github.com/berkeley-dsep-infra/datahub/issues/4913


ryanlovett (approved):
Looks good! Config is at hub.loadRoles as documented at https://z2jh.jupyter.org/en/latest/resources/reference.html and https://docs.datahub.berkeley.edu/en/latest/admins/howto/course-config.html#assigning-scopes-to-roles.

Issue #4934, opened 2023-08-25
Request more RAM for class Data 101


frank-lsf:
**Which hub do you want more RAM on? Please link the URL to the hub while specifying the hub name**

Data 101, https://data101.datahub.berkeley.edu/

**Which class is this request for?**

Data 101, fall 2023

**What is the Academic Guide URL for this class?**

https://www.data101.org/

**What is the bCourses URL for this class?**

https://bcourses.berkeley.edu/courses/1528038

**How many students do you expect in this class?**

300

**How much RAM per user is needed?**

4GB for course staff only.

**Why does this class need this much RAM?**

4GB for course staff is for developing course content and testing. Since we need to run Postgres, Mongo, etc. alongside the Jupyter notebooks, more RAM will be helpful.

**By what date (MM/DD) do you want this request to be fulfilled?**

8/31/2023

**Any additional information we should know about?**

We want a similar configuration as for Data 100: https://github.com/berkeley-dsep-infra/datahub/blob/524f721c8be626ce9bb2b7b7b4bd215baef805ae/deployments/data100/config/common.yaml#L63

cc @shaneknapp @balajialg

balajialg:
@franklsf95 Do you have an admin group with instructors for this particular course in bourses?

frank-lsf:
@balajialg yes, that would be "Teachers"

balajialg:
@ryanlovett Despite https://github.com/berkeley-dsep-infra/datahub/pull/4969, @franklsf95 can still not see the RAM increase. What might be going wrong here?

balajialg:
Closing this issue as this got resolved

PR #4948, opened 2023-08-28
Increase RAM to 4GB for Data 101 Admins

balajialg:
https://github.com/berkeley-dsep-infra/datahub/issues/4934

### DIFF
--- a/deployments/data101/config/common.yaml
+++ b/deployments/data101/config/common.yaml
@@ -19,6 +19,20 @@ jupyterhub:
         hub.jupyter.org/pool-name: core-pool-2023-07-11
 
   hub:
+    config:
+    loadRoles:
+      # Data 101, Fall 2023
+      course-staff-1528038:
+        description: Enable course staff to view and access servers.
+        # this role provides permissions to...
+        scopes:
+          - admin-ui
+          - list:users!group=course::1527426
+          - admin:servers!group=course::1527426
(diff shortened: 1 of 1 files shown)


deployments/data101/config/common.yaml
```diff
@@ -19,6 +19,20 @@ jupyterhub:
...
+        description: Enable course staff to view and access servers.
+        # this role provides permissions to...
+        scopes:
+          - admin-ui
+          - list:users!group=course::1527426
+          - admin:servers!group=course::1527426
+          - access:servers!group=course::1527426
+        # this role will be assigned to...
+        groups:
+          - course::1528038::group::Admins
```
ryanlovett, line 35: Looks good, as long as `course::1528038::group::Admins` exists. Perhaps double-check that the "Admins" bCourses group exists for the course?

ryanlovett (approved):

deployments/data101/config/common.yaml
balajialg: @ryanlovett I can't see a bcourses admin group for Data 101. I see that 9 folks are affiliated as a teacher. Does that mean the teachers are added to the admin group by default for this course? or should I inform instructors to create an admin group in bcourses?

balajialg: I will merge this PR and request the instructor to create a bcourses admin group (if needed).

PR #4969, opened 2023-08-31
Fixing issues with Data 101 Admin access

balajialg:


deployments/data101/config/common.yaml
```diff
@@ -193,6 +184,8 @@ jupyterhub:
       # https://bcourses.berkeley.edu/courses/1524699/groups#tab-80607
       course::1524699::group::all-admins:
         admin: true
+      course::1528038::group::all-admins:
```
ryanlovett: @balajialg `all-admins` is a group that exists in the bcourse 1524699, which is the "DataHub infrastructure" bcourse. There probably is no `all-admins` group in the bcourse 1528038 which is probably Data 101 for this term.

Courses don't need, and probably shouldn't have, admin=true. They should have the defined scopes above.

ryanlovett (commented):
Comment re: admin.

Issue #4985, opened 2023-09-01
Request more RAM for Data 100


mmiroyan:
**Which hub do you want more RAM on? Please link the URL to the hub while specifying the hub name**

Data 100 Hub (data100.hub.berkeley.edu)

**Which class is this request for?**

2023 Fall Data 100

**What is the Academic Guide URL for this class?**

https://classes.berkeley.edu/content/2023-fall-data-c100-001-lec-001

**What is the bCourses URL for this class?**

https://bcourses.berkeley.edu/courses/1527426

**How many students do you expect in this class?**

1110

**How much RAM per user is needed?**

4 GB

**Why does this class need this much RAM?**

One of the labs requires more than 2GB of RAM to run the notebook.

**By what date (MM/DD) do you want this request to be fulfilled?**

09/01/2023

**Any additional information we should know about?**

Can we please approve this request as soon as possible? Thanks.

mmiroyan:
https://github.com/berkeley-dsep-infra/datahub/pull/4986

felder:
This should land in prod in about 5 - 10m.

Admins with running servers will need to restart them to get the 4G guarantee.

PR #4986, opened 2023-09-01
lab 2 ram increase

mmiroyan:


felder:
https://github.com/berkeley-dsep-infra/datahub/issues/4985

PR #4996, opened 2023-09-06
adding content group

mmiroyan:


mmiroyan:
https://github.com/berkeley-dsep-infra/datahub/issues/4997

ryanlovett:
@mihranmiroyan All users currently have 4G/4G and this sets users in the new `::Content` group to 2G/4G. Is this what you want?

Setting admin users to 2G/4G does nothing since only tech infra staff are admins. Per slack with @felder I see that you're reverting a previous commit, but I'm just clarifying for other observers here.

mmiroyan:
updated the PR @ryanlovett

Issue #4995, opened 2023-09-06
Admin Access Request


dsconnell:
### Your Name

David Connell

### Affiliated Course Name

PHW251

### Bcourses Id

1525339

### Role

Instructional designer

### Name of the Hub

Public Health

### End Date

1/1/2024

balajialg:
@dsconnell I am closing this issue as the request got resolved but please reopen if course staff still don't have admin access

dsconnell:
Hi @balajialg I don't seem to have admin access? I've logged out and logged in, restarted, everything I can think to do.

balajialg:
@dsconnell Are you enrolled in this course as a teacher or TA (in bcourses)? Only folks enrolled in this course with these roles get admin access.

In addition, is this course published in bcourses? If not, please do publish it.

ryanlovett:
@balajialg It looks like @dsconnell 's role in the course is "designer" rather than "teacher", which we haven't seen before. David is this expected?

Should we give elevated access to all those who have that role for this course?

dsconnell:
Ah I didn't realize it was by role. I changed myself to "Teacher". The other designers should not have admin privileges on datahub.

balajialg:
Thanks @dsconnell for the clarity!

ryanlovett:
@dsconnell Okay, sounds good. In order for the hub to be aware of your new status in bCourses, you will need to logout and log back in. (not just stop/start server or close/open tab)

> The other designers should not have admin privileges on datahub.

As an aside, we scope the course staff roles so that people are no longer full hub admins. They only have the ability to administer the servers of those enrolled in the class. So we could give all designers the ability to do this if you wanted us to. If you assigning yourself to be Teacher in the course is consistent with your role and doesn't present any other challenges, then this is the simplest things for us too.

PR #5002, opened 2023-09-06
[DH-128] Add loadRoles config for PH 142 and PH 251 in Public Health Hub

balajialg:


balajialg:
Re: https://github.com/berkeley-dsep-infra/datahub/issues/4995

ryanlovett (approved):
Looks great!

Issue #4999, opened 2023-09-06
Request more RAM for class CE103N


cgerlein:
**Which hub do you want more RAM on? Please link the URL to the hub while specifying the hub name**

CE 103N Hub (https://datahub.berkeley.edu/user/cgerlein/tree/ce103nF2023-project)

**Which class is this request for?**

CE103N/ESPM130/GEOG136 

**What is the Academic Guide URL for this class?**

https://classes.berkeley.edu/content/2023-fall-civeng-c103n-001-lec-001

**What is the bCourses URL for this class?**

https://bcourses.berkeley.edu/courses/1527534

**How many students do you expect in this class?**

83

**How much RAM per user is needed?**

4GB

**Why does this class need this much RAM?**

The course includes 6 computer labs with multiple large datasets manipulation including meteorological data, satellite data, and mapping, all of which are memory-intensive 

**By what date (MM/DD) do you want this request to be fulfilled?**

09/06/2023

**Any additional information we should know about?**

First time using Datahub (both myself and GSI), we did not realize there was such a small limit on RAM. Lab #1 has been posted but the limit will make the notebook crash. We wanted to avoid having to set up all 83 students for Python on their machines and DataHub seemed like the best avenue for this.

balajialg:
@cgerlein I created a PR (#5003) to increase the RAM to 4 GB for your course and will share an update once it gets merged into the codebase. Couple of things to consider when you have time to make changes to the instructional workflow this semester or during the next semester,

1. Can you reduce the dataset size to fit the standard requirements? For eg: Using a subset of the dataset that is actually required for the assignment rather than the entire dataset (Highlighting this point as increased RAM correlates with increased cloud cost at our end)
2. There exists a Civil Engineering hub (https://cee.datahub.berkeley.edu) in case you want to use that for your coursework. It has QGIS installed in the image to support CIVENG 70 (if it is relevant to your course)

cgerlein:
Thanks, we are planning to reduce the dataset sizes that we had initially planned, but I'm expecting that some of the visualizations that are part of the labs will require the RAM. We hit the 1GB limit this week by asking the students to make a bar plot, which I did not expect to be so memory-intensive. We do not need QGIS but it might make sense for next year to set ourselves up in the CEE hub.

balajialg:
[...]

PR #5003, opened 2023-09-06
Increase RAM to 4GB for CE103N in Datahub

balajialg:
Ref: https://github.com/berkeley-dsep-infra/datahub/issues/4999


ryanlovett (approved):
Looks good!

PR #5027, opened 2023-09-13
fixing dependabot alerts

shaneknapp:
do we even still use this image for anything?  @ryanlovett @felder


ryanlovett:
No, there used to be a hub sidecar for this but it was disabled. Related code should be removed.

shaneknapp:
nothing makes me happier than deleting code!  i'll close this and make a new PR for removing this bit.

PR #5029, opened 2023-09-14
remove the fetch_course_emails sidecar

shaneknapp:
from @ryanlovett this isn't needed anymore and should be removed.


ryanlovett (approved):
Looks good!

PR #5060, opened 2023-09-22
install latest kubespawner on hub image

shaneknapp:
trying out @yuvipanda 's suggestion from https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/3229#issuecomment-1731958523

hash generation:
```
(dh) ➜  kubespawner git:(main) git rev-parse origin/main
d60146f5fe9cd31e09acf13c377d9334ecf59c9b
```


ryanlovett:
@shaneknapp Great that it has been fixed upstream! I think this change is probably sufficient.

The other approach would be to update the version in `hub/requirements.yaml` and `images/hub/Dockerfile` from 3.0.2 to, say, the version that was released today, https://hub.jupyter.org/helm-chart/#development-releases-jupyterhub. This might be necessary if kubespawner master needs other newer things that the chart brings in. But it might be worthwhile to test just the kubespawner update.

shaneknapp:
> @shaneknapp Great that it has been fixed upstream! I think this change is probably sufficient.
> 
> The other approach would be to update the version in `hub/requirements.yaml` and `images/hub/Dockerfile` from 3.0.2 to, say, the version that was released today, https://hub.jupyter.org/helm-chart/#development-releases-jupyterhub. This might be necessary if kubespawner master needs other newer things that the chart brings in. But it might be worthwhile to test just the kubespawner update.

yeah, let's just try kubespawner first...

shaneknapp:
ok, so i don't feel comfortable pushing this out until we're in off-hours.  this means i will either do that this weekend, or first thing monday morning.  tbh, probably the latter.

edit:  or perhaps first thing in the morning mid-week, so we can discuss a bit more on slack....

shaneknapp:
just tested on logodev-staging w/a couple of users and after a chartpress, the logs show a restart but no error messages regarding orphaned users.

i'll let the idle culler do it's thing and see if the pods end up  orphaned.

PR #5076, opened 2023-09-27
Update delete-unused-users.py

moschlar:
Fix logic error in to_delete selection

:wink:

### DIFF
--- a/scripts/delete-unused-users.py
+++ b/scripts/delete-unused-users.py
@@ -36,7 +36,7 @@ async def main():
                 except:
                     print(user['last_activity'])
                     raise
-                if last_activity and datetime.now().astimezone() - last_activity < timedelta(hours=24) and user['server'] is not None:
+                if last_activity and datetime.now().astimezone() - last_activity < timedelta(hours=24) or user['server'] is not None:
                     print(f"Not deleting {user['name']}")
                 else:
                     to_delete.append(user['name'])


shaneknapp:
thanks for the fix...  and this script definitely needs a refactor regardless.

i'm also curious how you stumpbled across this repo and script...  :)

moschlar:
@shaneknapp I searched for "JupyterHub delete inactive user" or something like that and stumbled upon your documentation page at https://docs.datahub.berkeley.edu/en/latest/admins/howto/remove-users-orm.html :wink: 

And if I were more familiar with async Python programming, I would have prepared a PR fixing the pagination issue - but my current hacky solution in jhub_client isn't really appropriate...

shaneknapp:
> @shaneknapp I searched for "JupyterHub delete inactive user" or something like that and stumbled upon your documentation page at https://docs.datahub.berkeley.edu/en/latest/admins/howto/remove-users-orm.html 😉
> 
> And if I were more familiar with async Python programming, I would have prepared a PR fixing the pagination issue - but my current hacky solution in jhub_client isn't really appropriate...

hah!  we are indeed the first result...  :)

async can be tricky, that's for sure.  since i inherited this script, i wanted to make it a little more human-parseable before trying to fix the pagination bug.

another way to do this can be found here:  https://discourse.jupyter.org/t/is-there-a-way-to-bulk-delete-old-users/20866/3

PR #5078, opened 2023-09-27
de-opacify the logic behind how we choose to delete users from a hub db

shaneknapp:
this script was a little confusing to parse, so i added a type check, additional output and cleaned up a confusing `if` statement.

### DIFF
--- a/scripts/delete-unused-users.py
+++ b/scripts/delete-unused-users.py
@@ -41,12 +41,11 @@ async def main():
                 except:
                     print(user['last_activity'])
                     raise
-
+                last_activity = asyncio
                 if isinstance(last_activity, datetime):
                     was_active_last_day = datetime.now().astimezone() - last_activity < timedelta(hours=24)
                 else:
-                    activity_type = type(last_activity)
-                    print(f"For user {user['name']}, expected datetime.datetime class for last_activity but got {activity_type} instead.")
+                    print(f"For user {user['name']}, expected datetime.datetime class for last_activity but got {type(last_activity)} instead.")
                     raise
 
                 print(f"User: {user['name']}")
--- a/scripts/delete-unused-users.py
+++ b/scripts/delete-unused-users.py
@@ -51,7 +51,7 @@ async def main():
                 print(f"Last login: {last_activity}")
                 print(f"24hrs since last login: {was_active_last_day}")
                 print(f"Running server: {user['server']}")
-                if (last_activity and was_active_last_day) or (user['server'] is not None):
+                if was_active_last_day or user['server'] is not None:
                     print(f"Not deleting {user['name']}")
                 else:
                     to_delete.append(user['name'])
--- a/scripts/delete-unused-users.py
+++ b/scripts/delete-unused-users.py
@@ -41,7 +41,6 @@ async def main():
                 except:
                     print(user['last_activity'])
                     raise
-                last_activity = asyncio
                 if isinstance(last_activity, datetime):
                     was_active_last_day = datetime.now().astimezone() - last_activity < timedelta(hours=24)
                 else:
--- a/scripts/delete-unused-users.py
+++ b/scripts/delete-unused-users.py
@@ -36,15 +36,21 @@ async def main():
                 except:
                     print(user['last_activity'])
                     raise
-                if last_activity and datetime.now().astimezone() - last_activity < timedelta(hours=24) or user['server'] is not None:
+                delta_bool = datetime.now().astimezone() - last_activity < timedelta(hours=24)
+                print(f"user: {user['name']}")
+                print(f"last login: {last_activity}")
+                print(f"24hrs since last login (bool): {delta_bool}")
+                print(f"server: {user['server']}")
+                if (last_activity and delta_bool) or (user['server'] is not None):
                     print(f"Not deleting {user['name']}")
                 else:
                     to_delete.append(user['name'])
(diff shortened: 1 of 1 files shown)


shaneknapp:
some sample output of a run (without deleting users, of course):
```
user: XXXX
last login: 2023-09-01 19:52:01.354394+00:00
24hrs since last login: False
server: None
Deleting XXXX

user: XXXX
last login: 2023-09-27 20:22:54.683580+00:00
24hrs since last login: True
server: /user/XXXX
Not deleting XXXX

user: XXXX
last login: 2023-09-27 17:30:56.352292+00:00
24hrs since last login: True
server: None
Not deleting XXXX

user: XXXX
last login: 2023-08-07 21:55:12.489883+00:00
24hrs since last login: False
server: None
Deleting XXXX

user: XXXX
last login: 2023-09-21 18:15:22.552542+00:00
24hrs since last login: False
server: None
Deleting XXXX

user: XXXX
last login: 2023-09-24 05:07:04.235891+00:00
24hrs since last login: False
server: None
Deleting XXXX

user: XXXX
last login: 2023-08-11 04:23:35.881508+00:00
24hrs since last login: False
server: None
Deleting XXXX

....

164 of 169: deleting XXXX
Skipped due to dry run.
165 of 169: deleting XXXX
Skipped due to dry run.
166 of 169: deleting XXXX
Skipped due to dry run.
167 of 169: deleting XXXX
Skipped due to dry run.
168 of 169: deleting XXXX
Skipped due to dry run.
169 of 169: deleting XXXX
Skipped due to dry run.
```

scripts/delete-unused-users.py
```diff
@@ -36,15 +41,24 @@ async def main():
                 except:
                     print(user['last_activity'])
                     raise
-                if last_activity and datetime.now().astimezone() - last_activity < timedelta(hours=24) or user['server'] is not None:
+                delta_bool = datetime.now().astimezone() - last_activity < timedelta(hours=24)
+                print(f"user: {user['name']}")
+                print(f"last login: {last_activity}")
+                print(f"24hrs since last login: {delta_bool}")
+                print(f"server: {user['server']}")
+                if (last_activity and delta_bool) or (user['server'] is not None):
```
shaneknapp: @ryanlovett @gmerritt could you look at the original `if` statement (like 39 of original file) and confirm i unpacked this properly?  my test runs lead me to believe i did, but another set of eyeballs would make me happier.

ryanlovett: If `last_activity` is ever something other than a datetime object (if `parse` returns None for example), then the `delta_bool` assignment will fail. 

I might also rename `delta_bool` to be `time_since_last_activity` or some such to make it a little quicker for human eyes to parse.

ryanlovett (changes requested):
Described a potential issue.

scripts/delete-unused-users.py
shaneknapp: i added some type checking logic, but looking at the docs for `dateutil.parser.parse` it should return a `datetime.datetime` object or raise an exception.  if it raises an exception, the script won't continue and we won't hit the condition you were worried about.

shaneknapp:
some testing output from the class check with `last_activity = 'asdfasdf'`:
```
For user XXXX, expected datetime.datetime class for last_activity but got <class 'str'> instead.
Traceback (most recent call last):
  File "/home/sknapp/src/datahub/scripts/./delete-unused-users.py", line 71, in <module>
    asyncio.run(main())
  File "/home/sknapp/miniconda3/envs/dh/lib/python3.10/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/home/sknapp/miniconda3/envs/dh/lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
    return future.result()
  File "/home/sknapp/src/datahub/scripts/./delete-unused-users.py", line 50, in main
    raise
RuntimeError: No active exception to reraise
```

and just to be sure, `last_activity = asyncio`:
```
For user XXXX, expected datetime.datetime class for last_activity but got <class 'module'> instead.
Traceback (most recent call last):
  File "/home/sknapp/src/datahub/scripts/./delete-unused-users.py", line 71, in <module>
    asyncio.run(main())
  File "/home/sknapp/miniconda3/envs/dh/lib/python3.10/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/home/sknapp/miniconda3/envs/dh/lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
    return future.result()
  File "/home/sknapp/src/datahub/scripts/./delete-unused-users.py", line 50, in main
    raise
RuntimeError: No active exception to reraise

File: scripts/delete-unused-users.py [hunk 2]
```diff
@@ -36,15 +41,30 @@ async def main():
                 except:
                     print(user['last_activity'])
                     raise
-                if last_activity and datetime.now().astimezone() - last_activity < timedelta(hours=24) or user['server'] is not None:
+                last_activity = asyncio
```
ryanlovett: I'm unclear on this since 1) it overrides the assignment in the `try` above and it will never represent the user's last activity, and 2) `asyncio` is a library.

ryanlovett (changes requested):
I think you can omit the `last_activity = asyncio` line? (was that for debugging?)

Also the `isinstance` check now means you can omit the check for `last_activity` in the new line 55. So it'd just be `if was_active_last_day or user['server'] is not None:`.

PR #5083, opened 2023-09-28
[DH-123] remove the pinned version of `kubespawner`

shaneknapp:
v6.1 was released today, which contains the fix for the orphaned resources and disconnection from a hub's underlying filesystem.

https://pypi.org/project/jupyterhub-kubespawner/
https://jupyterhub-kubespawner.readthedocs.io/en/6.1.0/changelog.html#id2


images/hub/Dockerfile
```diff
@@ -6,7 +6,6 @@ USER root
 
 RUN python3 -m pip install --no-cache git+https://github.com/berkeley-dsep-infra/canvasoauthenticator@09e7ed8
 RUN python3 -m pip install --no-cache jupyterhub-ltiauthenticator==1.5.1
-RUN python3 -m pip install --no-cache git+https://github.com/jupyterhub/kubespawner@main
```
ryanlovett, line 9: I think we will still need:
```
RUN python3 -m pip install --no-cache git+https://github.com/jupyterhub/kubespawner@6.1.0
```
Otherwise kubespawner will be downgraded to the version in our helm chart, which was broken.

ryanlovett (changes requested):
I believe removing the line will cause things to break again, and keeping `main` will work for a while, but tracking the latest git will eventually hurt us.

images/hub/Dockerfile
shaneknapp: got it!

ryanlovett:
Looks good!

shaneknapp:
alright, i'll chartpress locally and test on logodev before merging

shaneknapp:
testing seemed to show that everything worked!

PR #5093, opened 2023-09-29
[DH-145] fix pagination, basically a complete rewrite

shaneknapp:

PR #5095, opened 2023-10-02
Update remove-users-orm.rst

moschlar:
As mentioned in https://github.com/berkeley-dsep-infra/datahub/pull/5076#issuecomment-1741249455


shaneknapp:
thanks for the PR, but here's a PR for a rewrite of this tool using that discourse post as my template:

https://github.com/berkeley-dsep-infra/datahub/pull/5093

once i merge that, i'll share it w/the community/post a reply on discourse.  :)

PR #5104, opened 2023-10-05
add argument for inactive_since to specify other timedeltas 

moschlar:
Following up on https://github.com/berkeley-dsep-infra/datahub/pull/5076#issuecomment-1741249455, #5095 and #5093 

I added an argument to configure the inactivity period after which users should be considered for deletion.

(The advanced timedelta argument parsing was more of a fun exercise than really necessary :wink:)

(Also, I'd like to politely ask you to add the _hacktoberfest-accepted_ label to this PR if possible)

### DIFF
--- a/scripts/delete-unused-users.py
+++ b/scripts/delete-unused-users.py
@@ -22,7 +22,6 @@ import requests
 import sys
 
 from dateutil.parser import parse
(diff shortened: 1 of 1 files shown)


shaneknapp:
thanks for the PR!  i tested this locally and it behaves as expected.  :)

PR #5118, opened 2023-10-08
Pull {infra-,}requirements into environment.yml.

ryanlovett:
We want the dependencies to be resolved at the same time. Currently we install a batch with conda, then infra libs with pip, then various other libs with pip. The pip runs upgrade libraries installed by conda which is surprising if you're counting on the versions in environment.yml to be accurate. For example something in requirements.txt is upgrading numpy which is specified in environment.yml.

After splicing in the pip lists, I had to bump a few things to get everything to resolve. With this PR, the environment is what we're specifying. This could impact some courses that relied on libraries that were being inadvertently upgraded.


ryanlovett:
@shaneknapp @balajialg Please see comments above. This sets the libs to be what we thought they were, but they may be different than what was being installed up until now.

shaneknapp:
> @shaneknapp @balajialg Please see comments above. This sets the libs to be what we thought they were, but they may be different than what was being installed up until now.

yeah, wow.  this is pretty comprehensive, and also concentrates the scary in to `environment.yml`.  :)

these packages need a usage audit as there's a ton of stuff i see in there that obviously outdated (data 100/101/102 for example)...  i'm loathe to start removing packages until we get the python popcon package working.

ryanlovett:
@shaneknapp Yeah, I had to fight off the urge to remove data100, astro, etc. and all of the circa 2019 libraries until after the semester.

shaneknapp (approved):
LGTM....  🤞

balajialg (approved):

shaneknapp:
> @shaneknapp Yeah, I had to fight off the urge to remove data100, astro, etc. and all of the circa 2019 libraries until after the semester.

doing this will definitely give us some good feels, and i'd be happier fixing any issues that pop up if this breaks things for (what i hope will be) a relatively small group of instructors.

PR #5144, opened 2023-10-17
memory increase (project A2)

mmiroyan:


mmiroyan:
https://github.com/berkeley-dsep-infra/datahub/issues/5145

PR #5170, opened 2023-10-27
[DH-66] move conda and R install locations, fix popcon deployment path, install temp version of popcon

shaneknapp:
https://jira-secure.berkeley.edu/browse/DH-66

repo2docker is installing python to `/srv/conda`, and these Dockerfile-based images are deploying to `/opt/conda`.

since the majority of our deployments are following the repo2docker paradigm, let's update these hubs to do the same.

we will also be moving the popcon ipython startup script to `/srv/conda/...` so that it's in the proper ipython search paths.

### DIFF
--- a/deployments/a11y/image/environment.yml
+++ b/deployments/a11y/image/environment.yml
@@ -30,6 +30,7 @@ dependencies:
   # https://github.com/berkeley-dsep-infra/datahub/issues/3693
   - notebook==7.0.2
   - jupyterlab==4.0.4
+  - popularity-contest==git+https://github.com/yuvipanda/python-popularity-contest@9e30ec414e45615e75a0e86424141b7093c8f686
   # ###
   # The items below are from infra-requirements, however lab conflicts with the
   # alpha notebook.
--- a/deployments/a11y/image/environment.yml
+++ b/deployments/a11y/image/environment.yml
@@ -30,7 +30,7 @@ dependencies:
   # https://github.com/berkeley-dsep-infra/datahub/issues/3693
   - notebook==7.0.2
   - jupyterlab==4.0.4
-  - popularity-contest==git+https://github.com/yuvipanda/python-popularity-contest@9e30ec414e45615e75a0e86424141b7093c8f686
+  - git+https://github.com/yuvipanda/python-popularity-contest@9e30ec414e45615e75a0e86424141b7093c8f686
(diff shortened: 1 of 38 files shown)


shaneknapp:
re builds failing:

```
W: Failed to fetch https://cloud.r-project.org/bin/linux/ubuntu/focal-cran40/InRelease  Could not resolve 'cloud.r-project.org'
```

cloud.r-project.org is unreachable, as is r-project.org :(

ryanlovett:
This will help. A simpler way would be to set the popcon path to be under /etc/ipython which every ipython reads. But aligning our dockerfile builds with the others helps to avoid future surprises.

shaneknapp:
> This will help. A simpler way would be to set the popcon path to be under /etc/ipython which every ipython reads. But aligning our dockerfile builds with the others helps to avoid future surprises.

plan A was to stick the popcon script in `/etc` but after a few minutes of thought i decided that normalizing installation locations was more bueno than not.

shaneknapp:
> > This will help. A simpler way would be to set the popcon path to be under /etc/ipython which every ipython reads. But aligning our dockerfile builds with the others helps to avoid future surprises.
> 
> plan A was to stick the popcon script in `/etc` but after a few minutes of thought i decided that normalizing installation locations was more bueno than not.

actually, you're right and `/etc/ipython` is the right place for the startup script to live. 

https://github.com/berkeley-dsep-infra/datahub/pull/5170/commits/5e23ef5ebb05dbdf8806f29645a4693d759c0187

ryanlovett:
I think both changes will be helpful!

shaneknapp:
btw we're blocked on https://github.com/yuvipanda/python-popularity-contest/pull/3 for data101 and stat159 hubs (both have `ipydatagrid` installed).

shaneknapp:
welp this has been a bunch of work but i think we're good to go!

shaneknapp:
> btw we're blocked on [yuvipanda/python-popularity-contest#3](https://github.com/yuvipanda/python-popularity-contest/pull/3) for data101 and stat159 hubs (both have `ipydatagrid` installed).

i'm actually just using the git resolver for pip to get around this for now...  it doesn't work on a couple of hubs for random reasons but (thankfully) none of them have the `ipydatagrid` package installed.

PR #5174, opened 2023-10-30
[DH-163] fix julia image build

shaneknapp:


shaneknapp:
i bet this is why the build failed:  https://github.com/conda-forge/miniforge#whats-the-difference-between-mambaforge-and-miniforge

```
Before that release, Miniforge only shipped conda, while Mambaforge added mamba on top. Since Miniconda started shipping
conda-libmamba-solver in July 2023, Miniforge followed suit and started shipping it too in August. At that point, since 
conda-libmamba-solver depends on libmambapy, the only difference between Miniforge and Mambaforge was the presence of
the mamba Python package. To minimize surprises, we decided to add mamba to Miniforge too.
```

PR #5187, opened 2023-11-02
[DH-66] duplicate install of `jupyter-rsession-proxy` might be breaking things

shaneknapp:
`jupyter-rsession-proxy` is installed both via the dockerfile and environment.yml, only needs to happen once.  let's try environment.yml first!


ryanlovett:
@shaneknapp Yes, no need for python packages to be installed via Dockerfile, especially if they're on pypi or conda-forge.

PR #5188, opened 2023-11-02
Update syncthing to the latest stable version in Ischool hub image 

balajialg:
and primarily to rebuild the Ischool image so that the latest Stargazer package version 5.2.3 gets added as part of the image - https://jira-secure.berkeley.edu/browse/DH-170


shaneknapp (approved):
if it builds, lgtm

PR #5216, opened 2023-11-10
reduce RAM limit (project A2)

mmiroyan:


mmiroyan:
https://github.com/berkeley-dsep-infra/datahub/issues/5217

PR #5226, opened 2023-11-17
alphabetize builds in circleci

shaneknapp:


ryanlovett (approved):
Makes sense! While there are some deployment dependencies (e.g. some hubs use the main datahub image), the image builds and hub deploys should be completely separate.

At one point the hubs that used the main datahub image were also configured in this file to build the image, which would result in the image being built multiple times, but I think that was fixed a couple of months back. Any hubs using the datahub image (except for datahub itself) should not have their own build-image stanza.

shaneknapp:
> Makes sense! While there are some deployment dependencies (e.g. some hubs use the main datahub image), the image builds and hub deploys should be completely separate.
> 
> At one point the hubs that used the main datahub image were also configured in this file to build the image, which would result in the image being built multiple times, but I think that was fixed a couple of months back. Any hubs using the datahub image (except for datahub itself) should not have their own build-image stanza.

yeah, i am pretty sure that i'm not building hub images that are based on datahub.

PR #5231, opened 2023-11-22
[DH-134] Add more fields to performance issue template

balajialg:
Adding more inputs as discussed during the sprint planning meeting in October!


balajialg:
Tested the UI change in the Github preview and it is working!

PR #5238, opened 2023-11-28
[DH-182] critical security update for ray

shaneknapp:
there are three critical security issues in ray:
https://github.com/berkeley-dsep-infra/datahub/security/dependabot/73
https://github.com/berkeley-dsep-infra/datahub/security/dependabot/74
https://github.com/berkeley-dsep-infra/datahub/security/dependabot/75

we're on 1.3, we need to upgrade to at least 2.6.  i'll go for latest (2.8).

this is pretty bad, so i'll shepherd this through and get it out today.

@balajialg @ryanlovett @felder

### DIFF
--- a/deployments/datahub/images/default/environment.yml
+++ b/deployments/datahub/images/default/environment.yml
@@ -251,7 +251,7 @@ dependencies:
   - ipycanvas==0.9.0
 
   # data100 scientific packages
-  - ray==1.13.*
+  - ray==2.8.0
   - xlrd==2.0.1
 
   # data100 visualization


ryanlovett:
This would be bad if remote users could access the ray user's server over the network. I don't know anything about ray, but in z2jh it is challenging for one user to open a port on their server that another user could access. (like, I think the user would have to install (into their own account) and enable a jupyter server extension which required no authentication) I don't think our users are able to talk to the cluster controller to create new pods or other k8s resources.

So while the ray issue might be critical and would make sense to update, it could also be impossible to exploit on the hubs.

shaneknapp:
> This would be bad if remote users could access the ray user's server over the network. I don't know anything about ray, but in z2jh it is challenging for one user to open a port on their server that another user could access. (like, I think the user would have to install (into their own account) and enable a jupyter server extension which required no authentication) I don't think our users are able to talk to the cluster controller to create new pods or other k8s resources.
> 
> So while the ray issue might be critical and would make sense to update, it could also be impossible to exploit on the hubs.

i agree, though it is theoretically possible and could have potential impact in the form of an internal attack -- rather from the big scary internet at large.  theoretically possible, but for sure:  highly unlikely.  :)

anyways, it's critical and should be updated regardless.

PR #5248, opened 2023-11-30
[DH-135] install jupyterhub announcements

shaneknapp:


shaneknapp:
note:  https://z2jh.jupyter.org/en/latest/administrator/services.html

PR #5266, opened 2023-12-13
[DH-200] update the delete unused users script to accept a json config file

shaneknapp:


shaneknapp:
fwiw, i was worried that i might need to use `asyncio` for this, but the runtime for all of our hubs was only ~3 minutes!  XD

@ryanlovett

shaneknapp:
this has been tested against all live hubs using a config file, and single hubs using an env var + cli arg

scripts/delete-unused-users.py
```diff
@@ -15,22 +15,18 @@
...
 """
 import argparse
-from datetime import timedelta, datetime
+import json
 import logging
 import os
 import requests
 import sys
 
+from datetime import timedelta, datetime
```
ryanlovett: (To be pedantic) According to [PEP 8](https://peps.python.org/pep-0008/#imports) datetime should be listed with the other standard library imports.

scripts/delete-unused-users.py
```diff
@@ -15,22 +15,18 @@
...
 from dateutil.parser import parse
 
 logging.basicConfig(stream=sys.stdout, level=logging.WARNING)
 logger = logging.getLogger(__name__)
 
-token = os.environ["JUPYTERHUB_API_TOKEN"]
-headers = {
-    "Accept": "application/jupyterhub-pagination+json",
-    "Authorization": f"Bearer {token}",
-}
```
ryanlovett, line 33: It is interesting that the script was getting its authorization from the user's environment. This would work on the hub that the user was logged into, so I wonder why the script didn't also deduce the hub to operate on in the same way rather than having the user pass in the hub as an argument.

Being able to operate on multiple hubs is an improvement.

scripts/delete-unused-users.py
```diff
@@ -100,70 +96,115 @@ def should_delete(user, inactive_since):
...
 
-    count -= 1
-    print(f"Deleted {count} total users from the ORM.")
 
 if __name__ == "__main__":
     argparser = argparse.ArgumentParser()
+    argparser.add_argument(
+        '-f',
+        dest='credentials',
+        help='Path to a json file containing hub url and api keys.  Format is: {"hub1_url": "hub1_key", "hub2_url":, "hub2_key"}'
```
ryanlovett, line 182: The json format appears to use data as a key, which is concise. However I'd recommend that the keys not be data, e.g. `[{"url": "https://...", "key": "..."}, {"url": "https://...", "key": "..."}]`. Not strictly necessary of course, but good practice.

ryanlovett (approved):
Good idea! I left a couple of minor comments.

scripts/delete-unused-users.py
shaneknapp: but...  this way, the config is much more concise and looks cleaner/easier to read:

```
{
  "https://datahub.berkeley.edu": "XXXXXXXX",
  "https://logodev.datahub.berkeley.edu": "XXXXXXXX",
  "https://a11y.datahub.berkeley.edu": "XXXXXXXX",
  ...
}
```

fwiw, i started doing this your way but the amount of extra typing began to annoy me to no end.  ;)

scripts/delete-unused-users.py
shaneknapp: my understanding is that the script was initially meant to be run in a remote/non-hub environment.

YMMV.

shaneknapp:
script output w/o verbose or debug:

```
(dh) ➜  scripts git:(add-api-key-config) ✗ ./delete-unused-users.py -H https://prob140.datahub.berkeley.edu --dry_run
Scanning for users eligible for deletion on hub: https://prob140.datahub.berkeley.edu
Flagged 502 users for deletion.
Dry run:  Did not delete 502 total users from the ORM for hub: https://prob140.datahub.berkeley.edu
```

scripts/delete-unused-users.py
ryanlovett: Fair enough. For this little app it is fine, but you couldn't write a schema for this for example.

shaneknapp: noted!

PR #5274, opened 2023-12-15
[DH-140] clean up the datahub image python install

shaneknapp:
DO NOT MERGE (yet)

### DIFF
--- a/deployments/datahub/images/default/environment.yml
+++ b/deployments/datahub/images/default/environment.yml
@@ -1,13 +1,11 @@
 # https://docs.conda.io/projects/conda-build/en/latest/resources/package-spec.html#examples
 # For conda, == is exact constraint, while = is fuzzy constraint.
 # pip uses https://peps.python.org/pep-0440/ which does not have =.
-
 dependencies:
 - nodejs=16.*
 - traitlets=5.9.*
 - pip=22.2.*
 - python=3.9.*
-
 - syncthing==1.18.6
 
 # pymc3 needs this
@@ -76,10 +74,8 @@ dependencies:
 
 # data8; foundation
 - datascience==0.17.6
-
 - jupyter-server-proxy==3.2.1
 - jupyter-rsession-proxy==2.0.1
-
 - folium==0.12.1.post1
 
 # cogsci131; spring 2018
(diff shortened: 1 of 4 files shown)


deployments/datahub/images/default/environment.yml
```diff
@@ -250,51 +186,13 @@ dependencies:
...
-  - keras-preprocessing==1.1.2
-  - keras==2.11.0
-  - keras-vis==0.4.1
-  - plotly-express==0.4.1
-  - cytoolz==0.11.0
-  - pyro-ppl==1.7.0
-  - lime==0.2.0.1
-  - shap==0.39.0
-
   # prob140 2021 Spring
```
ryanlovett: We should either delete prob140 dependencies here, or preserve them, delete the prob140 hub, and have the prob140 instructor use datahub. The prob140 hub was created in order to isolate GSI access to student materials -- i.e. when data8 GSIs might be prob140 students, but this is now addressed by scopes and roles. Deleting the prob140 hub would require @balajialg / @ericvd-ucb to work with its course staff to ease the migration.

Deleting the dependencies here and preserving prob140 hub (or just doing nothing at this point) is the easier option, but I don't think its the right one long term.

ryanlovett (commented):
Yay for deleting a whole bunch of stuff for courses that now have their own hub!

I left one comment about prob140 dependencies.

Ideally, packages should be moved from the pip section to the upper conda package list, if those packages are available through conda. My feeling is that this will improve the performance of the resolver during installation.

Lastly, for dependencies with descriptions referring to very old terms (like Fall 2019), either those dependencies should be removed, or the comments should be updated to reflect that the dependencies are need through specific upcoming semesters. (like if it is known that the course will be taught by the same instructor for the next X semesters)

That being said, even if you agree with any of this, it doesn't all need to happen in this PR.

shaneknapp:
yeah, these are all great suggestions and i'll address as much as i can in
this PR.

the first thing i'd like to do is insert some timing to compare the overall
decrease in build time vs the original `environment.yml`.  sadly, i don't
think i'd be able to split out the conda vs pip section in there to see the
difference when moving things to the conda block.

deployments/datahub/images/default/environment.yml
shaneknapp: yep, you're right and we should eventually sunset the prob140 hub.  i'll comment these out and put a note in the comment.  i also just created https://jira-secure.berkeley.edu/browse/DH-203 to internally track this.

shaneknapp:
i ran the following command on a list of pip packages i gleaned from `environment.yaml`:

`for x in $(cat check-conda ); do conda search $x > /dev/null 2>&1 && echo $x >> in-conda; done`

then i went through, confirmed they were there and moved them out of the pip block:  https://github.com/berkeley-dsep-infra/datahub/pull/5274/commits/6d57c86a3f3a9bff960570d827f1197b4efc2fff

now we can check the mamba build steps w/timing and compare to how it was before!

shaneknapp:
tl;dr:  it didn't appear that having conda resolve more packages than pip would have been worth the effort to implement.  also, bumping the circleci runner size from 'default' to 'large' didn't appear to make the build go any quicker that before.

the final build is running and i'll be merging to staging afterwards.

PR #5277, opened 2023-12-15
[DH-140] time original datahub image build

shaneknapp:
in conjunction with #5274


shaneknapp:
got our data, large runner cut ~10-15m off of the build...

PR #5338, opened 2024-01-03
break out the apt R install steps for debugging/unsticking

shaneknapp:
see https://github.com/berkeley-dsep-infra/datahub/pull/5332

i was able to build the datahub image locally after breaking these apt-get installs out in to their own RUN blocks.  not sure why that helped, but we can at least test that out in circleci.


ryanlovett:
Sounds odd. I wonder if the base packages need to be ticked up. 

After it is debugged, it’s a good idea to have the apt update and install on the same RUN: https://docs.docker.com/develop/develop-images/instructions/#apt-get

shaneknapp:
> Sounds odd. I wonder if the base packages need to be ticked up.
> 
that's my guess...  but still only a guess.  :\

> After it is debugged, it’s a good idea to have the apt update and install on the same RUN: https://docs.docker.com/develop/develop-images/instructions/#apt-get

word

PR #5343, opened 2024-01-04
Change default URL for Data 8 hub to "/tree" and add entry for JUPYTERHUB_SINGLEUSER_APP

balajialg:
Changing the default URL from "/retro" to "/tree" based on docs - https://jupyter-notebook.readthedocs.io/en/v7.0.1/migrating/multiple-interfaces.html#nbclassic-and-notebook-7

### DIFF
--- a/deployments/data8/config/common.yaml
+++ b/deployments/data8/config/common.yaml
@@ -59,7 +59,7 @@ jupyterhub:
   singleuser:
     nodeSelector:
       hub.jupyter.org/pool-name: data8-pool
-    defaultUrl: /retro
+    defaultUrl: /tree
     storage:
       type: static
       static:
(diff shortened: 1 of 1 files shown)


ryanlovett:
Looks good.

If you want to run notebook with the new jupyter server app, you would need to install `nbclassic` and set `JUPYTERHUB_SINGLEUSER_APP` to `jupyter_server.serverapp.ServerApp` as well. (as [documented](https://z2jh.jupyter.org/en/stable/jupyterhub/customizing/user-environment.html#use-classic-notebook-by-default))

It might be good to transition away from the legacy notebook _server_ at some point (the app that supports notebook, lab, etc.), as we do in a11y, data100, data101, data102, edx, and stat159.

balajialg:
Thanks @ryanlovett! nbclassic package is already added to the Data 8 image. I just added the entry for JUPYTERHUB_SINGLEUSER_APP in the config file

balajialg:
I will merge this to staging for testing purposes and if something is off will revert the changes.

PR #5349, opened 2024-01-08
Refresh Data 8 image and test whether it launches notebook 7 and lab 4

balajialg:
List of tasks done as part of this PR,
-  Removed Dockerfile so that Data 8 follows repo2docker structure
- Added entries in the hubploy.yaml
- Commented out infra-reqs which provides old version of lab and notebook
- Added necessary packages from infra-reqs to env.yml file
- Removed packages under pip and moved them under conda block (if they had an entry in anaconda.org)

Thanks @ryanlovett for the suggestions!


shaneknapp (approved):
looks great!

ryanlovett (approved):
Looks good to me!

Issue #2992, opened 2021-11-12
Retrolab vs classic notebook - display of long arrays


davidwagner:
Long arrays are displayed differently in classic notebook vs Retrolab: they are put into a scrollable frame in classic notebook, but they spam the display in Retrolab.  For example, compare the output of cell 13 in lec10 demos:

classic notebook: http://data8.datahub.berkeley.edu/hub/user-redirect/git-sync?repo=https://github.com/data-8/materials-fa21&subPath=lec/lec10.ipynb

![lec10cell13classic](https://user-images.githubusercontent.com/3349841/141419886-d2a0a9f8-31b1-4316-9196-1aa305794391.png)

Retrolab: https://data8.datahub.berkeley.edu/hub/user-redirect/git-pull?repo=https%3A%2F%2Fgithub.com%2Fdata-8%2Fmaterials-fa21&branch=master&urlpath=retro%2Ftree%2Fmaterials-fa21%2Flec%2Flec10.ipynb

![lec10cell13retro](https://user-images.githubusercontent.com/3349841/141419897-df2e9cfa-0360-4eaf-8a08-6a928b8da22a.png)

In Retrolab the display of the array continues on for pages, without being truncated.  This is annoying.  I'm not sure there's an easy way to work around this in the notebook (by changing the cell).  What would you suggest?  Is there a way to get a more reasonable display for this?  Should I be filing this as an issue on the Retrolab github?

jtpio:
There is now an issue on the RetroLab repo: https://github.com/jupyterlab/retrolab/issues/287

And a WIP PR to fix it so the behavior is similar to the classic notebook: https://github.com/jupyterlab/retrolab/pull/288

As a workaround, explicitly setting `scrolled` in the cell metadata (https://nbformat.readthedocs.io/en/latest/format_description.html#cell-metadata) of the source material should automatically make the outputs scrollable in JupyterLab and RetroLab. We could imagine manually setting that `scrolled` flag for cells we know are going to generate long outputs.

yuvipanda:
Looks like this in retrolab now:

<img width="1357" alt="image" src="https://user-images.githubusercontent.com/30430/147960020-d58f2ce7-1a76-41e8-ab8a-aa20b9e4a8cc.png">

Thanks to @jtpio for fixing this!

PR #5351, opened 2024-01-09
Add retrolab package to the Data8 image

balajialg:
(Backup) In case the Data 8 instructional team decides to use the retro lab during SP 24!

### DIFF
--- a/deployments/data8/image/environment.yml
+++ b/deployments/data8/image/environment.yml
@@ -8,6 +8,7 @@ dependencies:
 - jupyter-resource-usage==1.0.1
 - nbgitpuller==1.1.0
 - syncthing==1.18.6
+- retrolab==0.3.21
 
 # data8
 - datascience==0.17.6


shaneknapp:
actually, do you need retrolab?  from their [github repo](https://github.com/jupyterlab/retrolab):

```
The RetroLab code base has now been integrated in the Jupyter Notebook GitHub repository: https://github.com/jupyter/notebook

Which means development is now happening in https://github.com/jupyter/notebook. Over the next weeks, we plan to triage the open issues in the RetroLab repository, and transfer them to the Notebook repo when relevant.
```

balajialg:
@shaneknapp Good point! RetroLab is incompatible with Lab 4.0 and the build failed for that particular reason. 

```
The following packages are incompatible
├─ jupyterlab 4.0.10  is requested and can be installed;
└─ retrolab 0.3.21  is uninstallable because it requires
   └─ jupyterlab >=3.4.0,<4 , which conflicts with any installable versions previously reported.
```

However, we need retro because I tried launching a new retro notebook in the Data 8 Staging hub which has the latest NB7 and Lab 4 and it returned a 404 error. Specifying retro in nbgitpuller prepends "retro/tree" to the URL

https://data8-staging.datahub.berkeley.edu/hub/user-redirect/git-pull?repo=https%3A%2F%2Fgithub.com%2Fds-modules%2FLegalst-123&branch=master&urlpath=retro%2Ftree%2FLegalst-123%2Flabs%2F03_Numpy+to+Pandas+and+Dataframe+Operations%2F03_Numpy_to_Pandas_and_Dataframe_Operations.ipynb

shaneknapp:
> However, we need retro because I tried launching a new retro notebook in the Data 8 Staging hub which has the latest NB7 and Lab 4 and it returned a 404 error. Specifying retro in nbgitpuller prepends "retro/tree" to the URL
> 
i'm not surprised that this is what's happening!  if retrolab isn't installed, then any links created via nbgitpuller that specify it will fail w/a 404.

basically, if retrolab is a requirement, we can't run jupyterlab 4.

is there a reason why we can't update the nbgitpuller links/docs/etc to reflect that retrolab is EOL?

balajialg:
@shaneknapp Yes, we need to update the nbgitpuller docs to reflect that retro lab is EOL. Some upstream engagement is required to remove RetroLab from this list - https://nbgitpuller.readthedocs.io/en/latest/link.html

Do you know if Notebook 7 and Lab 3 are compatible? Data 8 hub requires Nb 7 but doesn't need Lab 4. If we downgrade to Lab 3, we can support retro at least till spring break. Instructors need to regenerate the links for the course website and point them to the nb7 interface. Retro could act as a backup interface till the transition is done.

shaneknapp:
> @shaneknapp Yes, we need to update the nbgitpuller docs to reflect that retro lab is EOL. Some upstream engagement is required to remove RetroLab from this list - https://nbgitpuller.readthedocs.io/en/latest/link.html
> 
> Do you know if Notebook 7 and Lab 3 are compatible? Data 8 hub requires Nb 7 but doesn't need Lab 4. If we downgrade to Lab 3, we can support retro at least till spring break. Instructors need to regenerate the links for the course website and point them to the nb7 interface. Retro could act as a backup interface till the transition is done.

after confirming my suspicions locally w/repo2docker:  i don't think notebook 7 and jupyterlab < 4 will work together.

ryanlovett:
The instructors will need to update their nbgitpuller links anyways for the new [materials-sp24](https://github.com/data-8/materials-sp24) repo, so switching the links from retro to notebook at the same time makes sense.

I wonder if we can add a redirect on the user server from /retro/tree (retro) to /tree (nb7). Few would notice since NB7 _is_ the new retro. This can be done with a jupyter-server-proxy config, but it may be possible to do in a jupyter config or helm chart config. But it'd be much easier to ditch retro and communicate the change.

balajialg:
@ryanlovett I agree that switching to nb7 is the way forward. We might still need to support retro for a few weeks in case nb7 throws any new surprises for instructors. We encountered issues like this when we transitioned from classic to retro - https://github.com/berkeley-dsep-infra/datahub/issues/2992. nb7 has never been deployed in a course setting, unlike jl4 which was tested in Data 100 during SU 23.

A redirect for /retro/tree to /tree would be desirable but whether that is feasible given the tight timelines is a question I have.

balajialg:
Closing this PR as we decided not to worry about retrolab for the Data 8 hub!

PR #5360, opened 2024-01-10
Remove unnecessary entries from admin template

balajialg:
This stanza is no longer required given the current approach to providing admin access!


ryanlovett:
Merging, and then will follow up with another...

PR #5362, opened 2024-01-10
Add course-config.rst as part of docs

balajialg:
Seems like the course-config.rst file was deleted as part of this PR - https://github.com/berkeley-dsep-infra/datahub/pull/5283/files. Adding it back to our docs.

### DIFF
--- a/docs/admins/howto/course-config.rst
+++ b/docs/admins/howto/course-config.rst
@@ -1,0 +1,147 @@
+.. _howto/course-config:
+
+====================
+Course Configuration
+====================
+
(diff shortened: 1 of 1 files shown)


shaneknapp (dismissed):
~~please add an entry for this page in `docs/admins/howto/index.rst`~~

balajialg:
@shaneknapp There is already an entry here - https://github.com/berkeley-dsep-infra/datahub/blob/staging/docs/admins/howto/index.rst. I think this file was accidentally removed while the index entry remains untouched.

PR #5425, opened 2024-01-23
[edx] Added Package 88e: ipyleaflet

sean-morris:


sean-morris:
A quick package addition for edx.

PR #5429, opened 2024-01-23
Use dropdown for hub URL.

ryanlovett:


balajialg:
@ryanlovett Thanks a ton for doing this!

ryanlovett:
I'm not entirely sold its a great idea, since we'll have to update the issue template when we add/remove hubs, but it does increase the fidelity of new issues.

PR #5436, opened 2024-01-24
[DH-205] fix prom pod label to enable scraping

shaneknapp:
image

### DIFF
--- a/support/values.yaml
+++ b/support/values.yaml
@@ -95,7 +95,7 @@ prometheus:
       limits:
         cpu: 7
         memory: 48Gi
-    labels:
+    podLabels:
       # For HTTP access to the hub, to scrape metrics
       hub.jupyter.org/network-access-hub: 'true'
     persistentVolume:


ryanlovett:
@shaneknapp Did your successful HTTP request from the the prometheus server to the hub the other day indicate that  the network connection was not being blocked?

Either way though, it makes sense to use the proper syntax.

yuvipanda:
@ryanlovett he told me that it was through to the public DNS of the hubs, while prometheus uses kubernetes service discovery (https://prometheus.io/docs/prometheus/latest/configuration/configuration/#kubernetes_sd_config) to find out what to scrape. So it would hit the *hub pod IP* - the local 10.x IP - instead of the public DNS (which sidesteps networkpolicy completely).

https://github.com/2i2c-org/infrastructure/pull/2648 is the equivalent fix in the 2i2c infrastructure.

yuvipanda:
https://github.com/prometheus-community/helm-charts/blob/1f70a6b763a245f135f34c5a71f7149e3926d84c/charts/prometheus/values.yaml#L750 is the default prometheus helm chart's discovery config, so it makes it clearer *how* it generates the list of targets to scrape.

ryanlovett:
@yuvipanda Ah, makes sense, thanks!

PR #5438, opened 2024-01-24
[DH-205] set `interNamespaceAccessLabels` to allow the hubs to talk to prom

shaneknapp:
see also https://github.com/2i2c-org/infrastructure/pull/2648

also see also https://github.com/berkeley-dsep-infra/datahub/pull/5436

### DIFF
--- a/hub/values.yaml
+++ b/hub/values.yaml
@@ -227,6 +227,7 @@ jupyterhub:
       # hub.jupyter.org/network-access-hub=true label.
       #
       # ref: https://z2jh.jupyter.org/en/stable/resources/reference.html#hub-networkpolicy-internamespaceaccesslabels
+      # ref: https://github.com/2i2c-org/infrastructure/pull/2648
       #
       interNamespaceAccessLabels: accept
       ingress:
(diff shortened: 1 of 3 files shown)


hub/values.yaml
```diff
@@ -222,6 +222,13 @@ jupyterhub:
...
+      # interNamespaceAccessLabels=accept makes the hub pod's associated
+      # NetworkPolicy accept ingress from pods in other namespaces that has a
+      # hub.jupyter.org/network-access-hub=true label.
+      #
+      # ref: https://z2jh.jupyter.org/en/stable/resources/reference.html#hub-networkpolicy-internamespaceaccesslabels
+      #
+      interNamespaceAccessLabels: accept
       ingress:
         - from:
             - namespaceSelector:
```
shaneknapp, line 235: also perhaps we don't need the following block any more?  see https://github.com/2i2c-org/infrastructure/pull/2648/files#diff-301db94aa87a33a484be445796919615b3eecdc883900e976be6870e645f6d48L533

PR #5488, opened 2024-02-01
Adding shared directory to Econ 148 based on bcourses group affiliation

balajialg:
Just testing the functionality to have a shared/shread read write directories for a group based on bcourses affiliation.

@ericvd-ucb gave his approval to use Econ 148 as a guinea pig for this functionality.

Reviewed the documentation [here](https://docs.datahub.berkeley.edu/en/latest/admins/howto/course-config.html#defining-user-profiles) to make this change

bcourses group for reference,

image

### DIFF
--- a/deployments/data100/config/common.yaml
+++ b/deployments/data100/config/common.yaml
@@ -107,10 +107,10 @@ jupyterhub:
 
       # Econ148, Spring '24; testing shared_readwrite for groups in Eric's class
       course::1532866::group::shared_readwrite:
-        - mountPath: /home/jovyan/shared
-          name: home
-          subPath: _shared
-          readOnly: true      
+        extraVolumeMounts:
+          - name: home
+            mountPath: /home/jovyan/shared-readwrite
+            subPath: _shared    
     admin:
       mem_guarantee: 2G
       extraVolumeMounts:
--- a/deployments/data100/config/common.yaml
+++ b/deployments/data100/config/common.yaml
@@ -109,7 +109,7 @@ jupyterhub:
       course::1532866::group::shared_readwrite:
         extraVolumeMounts:
           - name: home
-            mountPath: /home/jovyan/econ148-readwrite
+            mountPath: /home/jovyan/_shared-econ148
             subPath: _shared    
     admin:
       mem_guarantee: 2G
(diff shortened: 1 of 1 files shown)


ryanlovett (commented):
I think it might need `extraVolumeMounts` as the dictionary key. (which the docs erroneously omit) For example:

```yaml
    course::1532866::group::shared_readwrite:
      extraVolumeMounts:
        - name: home
          mountPath: /home/jovyan/shared-readwrite
          subPath: _shared
```

Note that I've omitted `readOnly: true` since the comment says "shared_readwrite". I've also added `-readwrite` to the mount path. Our convention in datahub is to use `shared` for shared read-only directories and `shared-readwrite` for shared read-write directories.

You will also need to create the directory `_shared` in the filestore, along side the user home directories.

balajialg:
Thanks a ton, @ryanlovett! I replicated your suggestions in the config file. I might need some help with creating "_shared"directory in filestore. I will reach out via Slack!

balajialg:
@ryanlovett @shaneknapp How do I create a shared dir for this course in filestore? Is there any documentation that I can take a look at?

I logged into NFS server; Trying to find the location where homedirs are mounted.

ryanlovett:
@balajialg We don't have any concise info about just that, but we do have a few of the commands documented at https://docs.datahub.berkeley.edu/en/latest/admins/howto/new-hub.html#creating-a-new-filestore-instance.

Basically you should:
```bash
local$ gcloud compute ssh nfsserver-01 --zone=us-central1-b
nfsserver-01$ sudo install -d -o 1000 -g 1000 \
    /export/<hubname>-filestore/<hubname>/staging/_shared \
    /export/<hubname>-filestore/<hubname>/prod/_shared
```

We should probably have a separate section about this for creating shared data directories, right after the prod/shared directories are created.

balajialg:
Thanks a ton, @ryanlovett! Just made a commit that is related to your input

ryanlovett:
Since this is for Econ 148 on the Data 100 hub, I feel like it should be a separate shared directory, rather than a subdirectory of _shared. For example, `_shared-econ148`. Otherwise those in Data 100 can read/write data in Econ 148's directory. Maybe that's not a problem?

balajialg:
@ryanlovett Fair concern! Created a separate shared directory for Econ 148

ryanlovett:
@balajialg Okay, great. I also made one under data100/staging/.

balajialg:
Ah, missed that. Thanks @rylo! Can I merge this PR to staging and test it?

ryanlovett:
Yes, I think it is ready to test.

Issue #5515, opened 2024-02-06
Add gh-scoped-creds for dlab.datahub


aculich:
### Package Name

gh-scoped-creds

### Hub URL

dlab.datahub.berkeley.edu

### Course Name

For D-Lab workshops: https://github.com/dlab-berkeley/

Upcoming workshops: https://dlab.berkeley.edu/training/upcoming-workshops

### Semester Details

D-Lab workshops run year-round (including summer)

### Installation Deadline

Whenever it fits into your workload while also configuring it for other courses like @cboettig's ESPM 157, ESPM 288 in #5513

We have folks who will use it right away, and if there's anything about it you'd like D-Lab staff and grad students to help test, we'd be happy to before you implement it for other courses.

Likely we'll update our existing D-Lab workshop about using git, and will include how to use `gh-scoped-creds` in that workshop (see this past description): https://dlab.berkeley.edu/events/bash-git-introduction/2023-10-24

balajialg:
@aculich gh-scoped-creds is enabled in https://dlab-staging.datahub.berkeley.edu. It will be deployed to prod in the next few hours. Let us know how it works

cboettig:
this is _so_ nice!  My students are gonna love this.

balajialg:
@cboettig That's great! Once @aculich tests gh-scoped-creds and gives a thumbs up, I will replicate the same for your course

balajialg:
@aculich Closing this issue since the request got resolved via #5536. Please test the gh-scoped-creds workflow and reopen this issue if the are problems pushing to a git repository

cboettig:
@balajialg ah, ok we did discover one wrinkle here.  When students are using the RStudio interface, the `GH_SCOPED_CREDS_CLIENT_ID` environmental variable and `GH_SCOPED_CREDS_APP_URL` aren't visible on the terminal or the R session because RStudio insists on ignoring global env vars unless they are echoed into Renviron (e.g. into `$R_HOME/etc/Renviron.site`).  Is there a way to do this?

Currently students have to access this from the `/lab/` endpoint (which is fine with my grad students where we go back and forth between the jupyter hub and rstudio interface, but my undergrads don't know `/lab/`...)

balajialg:
[...]

PR #5536, opened 2024-02-13
[DH-231] Enable gh-scoped-creds for D-Lab

balajialg:
https://anaconda.org/conda-forge/gh-scoped-creds
https://github.com/berkeley-dsep-infra/datahub/issues/5515

### DIFF
--- a/deployments/datahub/images/default/environment.yml
+++ b/deployments/datahub/images/default/environment.yml
@@ -200,3 +200,7 @@ dependencies:
 
   # https://github.com/berkeley-dsep-infra/datahub/issues/5497
   - ottr-force-save-labextension==0.1.1
+
+
+  # [DH-231] Installing gh-scoped-creds for https://github.com/berkeley-dsep-infra/datahub/issues/5515
+  - gh-scoped-creds==4.1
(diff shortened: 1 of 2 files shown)


shaneknapp:
this looks good, but i have one question...  i went to the github app for this (https://github.com/apps/dlab-berkeley-datahub-access) and it seems that we're not done w/the configuration?

image

`Next: Confirm your installation location.`

i haven't added github scoped creds before, so i'm unsure if we're done w/setup and this is expected or not...  i did read the instructions but i don't want to start pushing random buttons and breaking things.  XD

https://github.com/jupyterhub/gh-scoped-creds?tab=readme-ov-file#github-app-configuration

balajialg:
@shaneknapp My understanding is that you need to install the Github app to the repo you want gh-scoped-creds to have access to. After that, you can push changes to that particular repository from the hub. Try installing the app in a personal Github repo?

I just enabled this app for a github repo below and the config seems to be fine (from an installation standpoint),

image

shaneknapp:
got it.  i set up my own app and went through the process.  thanks!

PR #5613, opened 2024-03-15
[DH-257] bump nbconvert, install chromium browser

shaneknapp:


balajialg:
@shaneknapp We might need to reopen this PR as @CharlieJCJ wants to fix the issue with `export notebook as webpdf` in Data 100 hub. Parallely, their content team will figure out whether they can fix latex issues which will allow the `export notebook as PDF` option to work

Issue #2677, opened 2021-09-02
Investigate data100 503 errors


yuvipanda:
From Andrew Lenz on Slack

> Data 100 is wondering if we can get more resources/computational power for DataHub. We constantly get overloaded for resources and get 503s on due dates, which is quite unfortunate.

There were many different issues provided, some which showed 503 and others exhibited themselves in other ways. We should track these, and fix them however we can. 

I don't think most of them are due to lack of resources though, except for some kernel deaths.

yuvipanda:
<img width="1422" alt="image" src="https://user-images.githubusercontent.com/30430/131911094-5bd95693-270a-4f31-91d6-0f1231c4e123.png">


Looking at server start times, I don't see anything too bad. I also don't see a lot of pending or waiting users

<img width="677" alt="image" src="https://user-images.githubusercontent.com/30430/131911299-70bfcfeb-5cef-413b-93f1-b501061140c1.png">

<img width="683" alt="image" src="https://user-images.githubusercontent.com/30430/131911342-8666fae4-a8d1-449a-982a-a28be8f13d02.png">

yuvipanda:
A bunch of different issues were reported!

![image](https://user-images.githubusercontent.com/30430/131921194-e9f0e9c6-8d45-47a7-bd31-eb3ee3cdfd37.png)

is one.  I haven't seen this one before, and there were several users reporting this. This is perhaps also a race condition between the multiple user schedulers we had, and could be fixed by #1817. Let's wait and watch to see if there are more reports of this.

balajialg:
Message from the teaching team - _sorry to tag you all (I know we decided on the Piazza route), but yet again, it's the due date for a major assignment and DataHub is down with a Service Unavailable screen (503?). Any ideas_

Link to [Piazza1](https://piazza.com/class/kqsiwfz12g0482?cid=445) and [Piazza 2](https://piazza.com/class/kqsiwfz12g0482?cid=439)where students raised the issue.

Request is to provide more resources to avoid such scenarios in the future!

felder:
@yuvipanda while I cannot prove this conclusively, there appears to be a V shaped currently running users graph here:
https://grafana.datahub.berkeley.edu/goto/g_YrOXSnz

With a correlation to this particular PR:
https://github.com/berkeley-dsep-infra/datahub/pull/2768

[...]

PR #5614, opened 2024-03-15
[DH-3] increase cpu allocation for data8 hub and proxy pods

shaneknapp:
data100 hasn't had anywhere nearly as many 503s as data8, and i figured that we could replicate the data100 prod settings and see if that helps.

ref:  https://github.com/berkeley-dsep-infra/datahub/issues/2677


ryanlovett (approved):
Makes sense! The comments still reference data101 so maybe replace "Give data101" with "Allocate" or some such.

PR #5650, opened 2024-04-01
Test whether google groups related info can be pulled to Datahub!

balajialg:
Not sure if this will work! Based on conversation with @ryanlovett in https://uctech.slack.com/archives/C04NEF48SCR/p1712003256681759


balajialg:
Closing this PR as we are not proceeding with this approach!

PR #5669, opened 2024-04-10
Update common.yaml

andrewpbray:
Updating the image for the gradebook hub


ryanlovett:
I confirmed the tag at https://hub.docker.com/r/gradebook/gradebook/tags.

PR #5686, opened 2024-04-19
[DH-274] Changing shared-readwrite from /opt to /home/jovyan in Stat 20 hub for infra admins!

balajialg:
Okay, I just changed the stanzas to mimic Stat 159 hub. Let me know if you are okay with it?

Re: https://github.com/berkeley-dsep-infra/datahub-usage-analysis/pull/17
https://jira-secure.berkeley.edu/browse/DH-274

### DIFF
--- a/deployments/stat20/config/common.yaml
+++ b/deployments/stat20/config/common.yaml
@@ -72,7 +72,7 @@ jupyterhub:
         subPath: "{username}"
       extraVolumeMounts:
         - name: home
-          mountPath: /home/jovyan/shared
+          mountPath: /opt/shared
           subPath: _shared
           readOnly: true
     memory:
@@ -86,9 +86,14 @@ jupyterhub:
       course::1524699::group::all-admins:
         admin: true
         mem_limit: 8G
+      # Shared read directory for infra admins to store logs data from every semester
+      course::1534506::enrollment_type::teacher:
+        extraVolumeMounts:
+          - name: home
+            mountPath: /home/jovyan/admins-shared-storage
+            subPath: _shared/admins-storage
     admin:
       extraVolumeMounts:
         - name: home
-          mountPath: /home/jovyan/shared-readwrite
+          mountPath: /opt/shared-readwrite
           subPath: _shared
-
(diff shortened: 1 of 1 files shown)


ryanlovett (changes requested):
First, sorry for not mentioning this earlier in the other thread where I referenced stat159 and /opt...

Second, I think we shouldn't alter the existing volume mounts -- we should add a new one just for this project. The existing ones are for course data. If you were to merge as is, and drop the data into `_shared`, the students would be able to see the data.

- This should be a new volume mount scoped under `group_profiles.course::1524699::group::all-admins:'
- It shouldn't use the directory name `shared-readwrite` since 1) the volume will be mounted read-only and 2) `shared-readwrite` and `shared` are part of a convention used in the hubs across all courses. That name should be set aside for course data. Since this change is for datahub admins, we can mount it as `/home/jovyan/shared-524699-admins` or some such.
- `_shared` should be `_shared-524699-admins` or some such since `_shared` is again reserved for course data.

balajialg:
Thanks @ryanlovett for the clarification. I was aware of the protocol for shared/shared-read-write directories since I had handled [such requests in the past](https://github.com/berkeley-dsep-infra/datahub/pull/5644/commits/31f2199e52b57bc0c9a6f93293d5e18cbbdafabd). However, I got confused with the admin stanza listed below. For a moment, I thought that this stanza needs to get changed for admins instead of adding a new stanza linking bcourses id (as we do for user requests). My bad! 

```
    admin:
      extraVolumeMounts:
        - name: home
          mountPath: /opt/shared-readwrite
          subPath: _shared
```

balajialg:
@ryanlovett @shaneknapp Let me know if you are okay with merging the changes to staging?

deployments/stat20/config/common.yaml
```diff
@@ -86,8 +86,14 @@ jupyterhub:
...
+        extraVolumeMounts:
+          - name: home
+            mountPath: /home/jovyan/admins-shared-storage
+            subPath: _shared/admins-storage
     admin:
       extraVolumeMounts:
         - name: home
           mountPath: /opt/shared-readwrite
-          subPath: _shared
+          subPath: _shared
```
shaneknapp: add a new line after `_shared` please

balajialg: Done

shaneknapp:
looks good to me...

PR #5690, opened 2024-04-23
Update Stat 20 R Packages

andrewpbray:
pull stat20data from github instead of via symlink


andrewpbray:
Interesting, I'm not seeing the change at https://stat20-staging.datahub.berkeley.edu/user/andrewbray/rstudio/

This was meant to pull in https://github.com/stat20/stat20data/commits/master/ with a new data set called `fastfood1`.

```r
library(stat20data)
fastfood1
```
```
Error: object 'fastfood1' not found
```
Does anything pop out to you as obviously wrong?

ryanlovett:
I think you were just a little quick on the draw. :) Once the PR is merged, it has to work its way through CI again (another image build) and then the image actually gets deployed to the hub. Also, if your server was started with the old configuration, you have to restart it before you see the change. This can be done at stat20-staging.datahub.berkeley.edu/hub/home.

I was able to successfully run the code when I tried a few moments ago in a new server.

andrewpbray:
bingo. good catch.

we have one more dataset to add so I've done one more package update PR.

PR #5693, opened 2024-04-24
[Test] downgrade nbgitpuller and jupyterhub in dev-r hub image

balajialg:
@shaneknapp Let me know if you are okay with making changes to infra-req for dev-r? I will test this in staging and then revert the changes. Yet another data point that we can use to report upstream.

### DIFF
--- a/deployments/dev-r/images/default/infra-requirements.txt
+++ b/deployments/dev-r/images/default/infra-requirements.txt
@@ -13,7 +13,7 @@ jupyterlab==4.0.11
 nbgitpuller==1.1.1
 jupyter-resource-usage==1.0.1
 # Matches version in images/hub/Dockerfile
-jupyterhub==4.0.1
+jupyterhub==4.0.2
 appmode==0.8.0
 ipywidgets==8.0.7
 jupyter-tree-download==1.0.1
--- a/deployments/dev-r/images/default/infra-requirements.txt
+++ b/deployments/dev-r/images/default/infra-requirements.txt
@@ -10,10 +10,10 @@
 # FIXME: Freeze this to get exact versions of all dependencies
 notebook==7.0.7
 jupyterlab==4.0.11
-nbgitpuller==1.2.1
+nbgitpuller==1.1.1
 jupyter-resource-usage==1.0.1
 # Matches version in images/hub/Dockerfile
-jupyterhub==4.1.5
+jupyterhub==4.0.1
 appmode==0.8.0
 ipywidgets==8.0.7
 jupyter-tree-download==1.0.1
--- a/deployments/dev-r/config/staging.yaml
+++ b/deployments/dev-r/config/staging.yaml
@@ -3,6 +3,10 @@ nfsPVC:
     shareName: shares/dev-r/staging
 
 jupyterhub:
+  hub:
+    image:
+      name: gcr.io/ucb-datahub-2018/jupyterhub-hub
+      tag: 0.0.1-0.dev.git.7126.h64800628
   scheduling:
     userScheduler:
       replicas: 1


shaneknapp:
this won't work as expected due to the hub image itself having a higher version of jupyterhub installed.  you'll need to build and push a new image w/the lower deps installed.

i believe that you can specify this in `staging.yml` as `hub.image.<etc>`.  @ryanlovett have you done this before?

shaneknapp (commented):
this won't work as expected due to the hub image itself having a higher version of jupyterhub installed.  you'll need to build and push a new image w/the lower deps installed.

i believe that you can specify this in `staging.yml` as `hub.image.<etc>`.  @ryanlovett have you done this before?

shaneknapp:
i'd also considering experimenting by also lowering the jupyterhub version to 4.0.x and keeping nbgitpuller at 1.2.1

balajialg:
@shaneknapp Got it, thanks. If you can redirect me to a documentation then I will take a look at it. Or else we can probably do this during a short check-in if you are available.

ryanlovett:
@shaneknapp I haven't done this exact thing before, but it should be possible to set a value in a deployment that masks any corresponding value in hub/values.yml. This would indeed be hub.image. To build a new hub image you'd modify images/hub/Dockerfile and build a new hub image with chartpress.

chartpress builds and pushes a new image, but it also changes the hub version in the top-level config automatically. So either you would let it do that, test a hub with hubploy, and not commit those changes until you want to deploy them for all hubs; or you would remove the top-level config changes to hub.image and manually insert them into the testing hub config, and commit that.

shaneknapp:
> @shaneknapp I haven't done this exact thing before, but it should be possible to set a value in a deployment that masks any corresponding value in hub/values.yml. This would indeed be hub.image. To build a new hub image you'd modify images/hub/Dockerfile and build a new hub image with chartpress.
> 
> chartpress builds and pushes a new image, but it also changes the hub version in the top-level config automatically. So either you would let it do that, test a hub with hubploy, and not commit those changes until you want to deploy them for all hubs; or you would remove the top-level config changes to hub.image and manually insert them into the testing hub config, and commit that.

ah i see.  i created a PR to do that...  could you take a look and we can carry that convo on over there?  https://github.com/berkeley-dsep-infra/datahub/pull/5696

shaneknapp:
> > @shaneknapp I haven't done this exact thing before, but it should be possible to set a value in a deployment that masks any corresponding value in hub/values.yml. This would indeed be hub.image. To build a new hub image you'd modify images/hub/Dockerfile and build a new hub image with chartpress.
> > chartpress builds and pushes a new image, but it also changes the hub version in the top-level config automatically. So either you would let it do that, test a hub with hubploy, and not commit those changes until you want to deploy them for all hubs; or you would remove the top-level config changes to hub.image and manually insert them into the testing hub config, and commit that.
> 
> ah i see. i created a PR to do that... could you take a look and we can carry that convo on over there? #5696

figured it out! https://github.com/berkeley-dsep-infra/datahub/pull/5696#issuecomment-2076026761

deployments/dev-r/images/default/infra-requirements.txt
```diff
@@ -10,10 +10,10 @@
 # FIXME: Freeze this to get exact versions of all dependencies
 notebook==7.0.7
 jupyterlab==4.0.11
-nbgitpuller==1.2.1
+nbgitpuller==1.1.1
 jupyter-resource-usage==1.0.1
 # Matches version in images/hub/Dockerfile
-jupyterhub==4.1.5
+jupyterhub==4.0.1
```
shaneknapp: change this to `4.0.2`, which matches the version on the dev-r staging hub.

shaneknapp (changes requested):

deployments/dev-r/images/default/infra-requirements.txt
balajialg: done

shaneknapp (approved):

shaneknapp:
i would suggest that after merging this and testing, to open another PR to bump nbgitpuller to `1.2.1` .

balajialg:
@shaneknapp Sounds good. Will do

Issue #5643, opened 2024-03-27
Demography Workshop Datahub Request


wrathofquan:
# Summary

We have two workshop events coming up in the Demography/Population sciences department and are looking at using the workshop hub or potentially another dedicated hub. 

- The first event is a workshop/hacakthon type event that actually will run for the entire summer and involves Berkeley faculty, graduate students and post docs. It's a data challenge called the [predicting fertility data challenge](https://stulp.gmw.rug.nl/prefer/). It begins April 1 and expect about 5-6 users training ML models in python and R. Since this is a longer running event (continues through summer), I'm not sure if the workshop hub is suitable but wanted to check in to see if you have suggestions or alternatives.

- This year the Demography department will again host a [week-long workshop, June 3-7](https://populationsciences.berkeley.edu/population-center/annual-workshop-on-formal-demography/#:~:text=The%209th%20Annual%20Workshop%20on,by%20the%20Berkeley%20Population%20Center.) on statistical methods in June with researchers visiting from all over the world. We used workshop hub last year with great success and hope to use it again! Compute needs will be same as last year - 4GB of RAM per user. We expect 20-40 users. 

# User Stories

Workshop instructors and faculty can have all of their instructional materials in a Datahub that is consistent across users.
Attendees won't have to worry about managing their own compute environments.

# Acceptance criteria

- For the data challenge event, users with a calnet can navigate to datahub and use python/r to train ml models. Users will be using same datasets so a `shared-read-write` directory would be required. 

- For the June workshop event, users can navigate to something like: workshop.datahub.berkeley.edu, authenticate without calnet (and potentially manage their own credentials), access the workshop files in RStudio or native R kernel.

# Important information

### Data challenge
- The data challenge (April 1 - September) will have about 5-6 users each needed about 4GB of RAM and a `shared-read-write` to work on shared datasets that exceed storage capacity in github. 

### Demography workshop
- This workshop is in June (3-7)  and we expect somewhere between 20-50 users
- We'd like to have a 4GB allocation of RAM per user.
- We'd want to keep the pods alive for a few weeks after the workshop to give users a chance to save their work.

# Tasks to complete

balajialg:
@wrathofquan Thanks, Responding to the time sensitive request first - Yes, students can use the workshop hub to do their work. We can enable shared-readwrite directories so that they can store their datasets. Is it possible to create a dummy bcourses site, add all the participants of the data challenge to that site and share the bcourses id? 

[...]

PR #5711, opened 2024-05-03
Increase RAM to 4 GB for workshop hub

balajialg:
To be merged during the last week of May. 

As requested in https://github.com/berkeley-dsep-infra/datahub/issues/5643. Creating this PR now so that it doesn't slip through the cracks.


balajialg:
This needs to get merged as the workshop starts June 3rd. Let's give the instructors few days to test the notebooks with increased RAM.

PR #5714, opened 2024-05-06
update repo2docker version pin

shaneknapp:


ryanlovett:
@shaneknapp You don't want to set it to [2024.3.0](https://pypi.org/project/jupyter-repo2docker/)? I worry about having it unversioned.

shaneknapp:
> @shaneknapp You don't want to set it to [2024.3.0](https://pypi.org/project/jupyter-repo2docker/)? I worry about having it unversioned.

fair point.  :)

PR #5731, opened 2024-05-15
Tell prometheus to use cloned disk containing data from previous cluster

felder:
https://jira-secure.berkeley.edu/browse/DH-278

### DIFF
--- a/support/values.yaml
+++ b/support/values.yaml
@@ -102,6 +102,7 @@ prometheus:
     persistentVolume:
       size: 1000Gi
       storageClass: ssd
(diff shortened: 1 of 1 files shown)


felder:
@shaneknapp this has already been pushed to the support chart manually via helm from my workstation.  I believe it worked, but we should double check before merging.

shaneknapp (approved):
lgtm!  confirming now...

shaneknapp:
i had to redeploy the jupyterhub dashboards and the metrics reappeared as expected!  i will manually update the custom panels we have today.

PR #5765, opened 2024-05-30
[DH-262] add PDF export and remove QtPNG export from datahub and r hub

felder:
https://jira-secure.berkeley.edu/browse/DH-262


felder:
Waiting for @balajialg to test, please do not merge.

PR #5767, opened 2024-05-31
[DH-299] Install jupyterlab extension published in pypi

balajialg:
Since https://github.com/berkeley-dsep-infra/datahub/pull/5743 didn't work. Trying a new approach to install the extension in the hub (thanks @ryanlovett)

### DIFF
--- a/deployments/a11y/image/postBuild
+++ b/deployments/a11y/image/postBuild
@@ -8,8 +8,8 @@ npm install -g pa11y@6.2.3
 
 # Install Node.js and Yarn for DH-299
 echo "Installing Node.js and Yarn..."
-sudo apt-get update && sudo apt-get install -y nodejs npm
-sudo npm install -g yarn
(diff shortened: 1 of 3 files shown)


deployments/a11y/image/postBuild
```diff
@@ -5,3 +5,35 @@ set -eux
 
 npm install -g pa11y-ci@3.1.0
 npm install -g pa11y@6.2.3
+
+# Install Node.js and Yarn for DH-299
+echo "Installing Node.js and Yarn..."
+apt-get update && apt-get install -y nodejs npm
```
ryanlovett: I don't think this can run as root. You might want to install nodejs, npm, yarn, etc. through conda.

It'd probably be quicker to nail down these steps in a live environment on the hub rather than iterating through CI.

balajialg: Thanks @ryanlovett! Will do

PR #5773, opened 2024-06-03
Install git from conda-forge.

ryanlovett:
Jammy's git is too old to support gh-scoped-creds, so install from conda-forge.


shaneknapp:
LGTM!

PR #5812, opened 2024-06-20
[DH-314] Install gh-scoped-creds app in a profile in dev hub 

balajialg:
Testing the configuration in one of the profiles in dev hub

### DIFF
--- a/deployments/dev/config/common.yaml
+++ b/deployments/dev/config/common.yaml
@@ -69,6 +69,9 @@ jupyterhub:
         default: true
       - display_name: "Dashboard image"
         description: "A newer repo2docker-based image with a primary focus of building dashboards."
+      extraEnv:
+        GH_SCOPED_CREDS_CLIENT_ID: "Iv23liFjgKFgkSBWph4C"
+        GH_SCOPED_CREDS_APP_URL: "https://github.com/apps/test-gh-scoped-creds"
         kubespawner_override:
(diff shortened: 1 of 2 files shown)


deployments/dev/config/common.yaml
```diff
@@ -69,6 +69,9 @@ jupyterhub:
         default: true
       - display_name: "Dashboard image"
         description: "A newer repo2docker-based image with a primary focus of building dashboards."
+      extraEnv:
```
ryanlovett, line 72: @balajialg It looks like the env is not indented the same as one of the profile stanzas. Also, I think `extraEnv` is not evaluated at the level of child of [profile_list](https://jupyterhub-kubespawner.readthedocs.io/en/latest/spawner.html#kubespawner.KubeSpawner.profile_list). It might have to be `extra_env` and as a child of `kubespawner_override`, if you wanted it to be specific to one of the profiles.

ryanlovett (commented):
@balajialg This might explain why the variables aren't being set in the env.

deployments/dev/config/common.yaml
balajialg: Ah, got it. Thanks @ryanlovett. I will try to fix it in the next PR

balajialg: Yes, I wanted it to be the child of kubespawner_override

PR #5829, opened 2024-07-01
[DH-317] Install quarto in Dev hub (dashboard image)

balajialg:
So that Sophie can explore quarto dashboards in Dev hub


ryanlovett:
I recommend not using R code to script the installation of quarto. I also think this method will fail because I believe the user running install.R in repo2docker doesn't have root privileges. I believed this passed CI because it is the secondary image and hubploy doesn't build secondary images.

repo2docker doesn't make quarto installation easy because quarto is not in Ubuntu jammy repos ~~nor is it in conda-forge~~.
Upstream's example at https://github.com/cwickham/binder-example/ installs into ~/.local/bin, but ~ gets displaced by users' hub storage. 

~~So either we have to use a Dockerfile, like what stat20 uses, or we have to find a way to seed content for [repo2docker's appendix](https://repo2docker.readthedocs.io/en/latest/usage.html#cmdoption-jupyter-repo2docker-appendix). That can run commands as root.~~

Lastly, it'd probably worth creating an issue to request that quarto be installed (along with RStudio, shiny, etc.) in repo2docker's R build packs.

ryanlovett:
Actually, there *is* a quarto package in conda, https://anaconda.org/conda-forge/quarto. So it should just go into the environment.yml.

balajialg:
Thanks @ryanlovett for the suggestion. I just made a commit which added the quarto package to the environment.yml file. That seems to be a simple process to enable quarto. Good to know!

deployments/dev/images/secondary/install.R
```diff
@@ -29,4 +30,4 @@ cran_packages = c(
   "shiny", "1.7.4"
 )
 
-class_libs_install_version("Base packages", cran_packages)
+class_libs_install_version("Base packages", cran_packages)
```
shaneknapp: add a newline pls

balajialg: Done

deployments/dev/images/secondary/install.R
```diff
@@ -1,5 +1,6 @@
-#!/usr/bin/env r
+#!/usr/bin/env R
```
ryanlovett, line 1: I was wondering about this since it is technically correct, but hasn't been a problem. This script is evidently passed to R rather than made executable and run directly. The [upstream example](https://github.com/binder-examples/r/blob/HEAD/install.R) doesn't even have a hashbang.

balajialg:
@ryanlovett I will merge the PR for now. Will fix it if there are issues.

PR #5832, opened 2024-07-02
address dependabot issues

shaneknapp:
https://github.com/berkeley-dsep-infra/datahub/security/dependabot/184
https://github.com/berkeley-dsep-infra/datahub/security/dependabot/185
https://github.com/berkeley-dsep-infra/datahub/security/dependabot/186


shaneknapp:
i will merge this to `staging` tomorrow and do some testing.  if everything seems to be ok, i will merge to `prod`.

PR #5840, opened 2024-07-03
[DH-227] i ran yamllint and fixed all of the pertinent errors

shaneknapp:
weirdly it's reporting errors in directories that i explicitly tell it to ignore:  `**/template/**` and `**/templates/**`...

anyways.


deployments/data101/config/common.yaml
```diff
@@ -91,21 +91,6 @@ jupyterhub:
         c.Spawner.pre_spawn_hook = ensure_db_pvc
     nodeSelector:
       hub.jupyter.org/pool-name: core-pool-2024-05-08
-    config:
```
shaneknapp, line 94: @balajialg this block of code was duplicated, and is defined higher up in the config

deployments/data8/config/common.yaml
```diff
@@ -5,6 +5,11 @@ nfsPVC:
 
 jupyterhub:
   custom:
+    group_profiles:
```
shaneknapp, line 8: the `custom:` key was duplicated, so i moved the contents of the lower block up here.

deployments/data101/config/common.yaml
balajialg: Ah, got it. Thanks for the fix

PR #5841, opened 2024-07-03
add postBuild to the Dockerfile

shaneknapp:
you need to explicitly add postBuild to the Dockerfile for repo2docker to know to build it.

@balajialg 

i tested this out on my desktop and the build will fail with `Error: unknown flag: --data-dir`.

### DIFF
--- a/deployments/dev/images/default/Dockerfile
+++ b/deployments/dev/images/default/Dockerfile
@@ -125,7 +125,11 @@ RUN R -e "IRkernel::installspec(user = FALSE, prefix='${CONDA_DIR}')"
 
 # clear out /tmp
 USER root
+COPY postBuild /tmp/postBuild
+RUN chmod +x /tmp/postBuild
+RUN /tmp/postBuild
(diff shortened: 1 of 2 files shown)


balajialg:
Ah, thanks a lot @shaneknapp! I didn't know that. I will look into it the reported error now

shaneknapp:
fyi that flag doesn't seem to exist:

```
(dh) ➜  incoming ./ollama run --help
Run a model

Usage:
  ollama run MODEL [PROMPT] [flags]

Flags:
      --format string      Response format (e.g. json)
  -h, --help               help for run
      --insecure           Use an insecure registry
      --keepalive string   Duration to keep a model loaded (e.g. 5m)
      --nowordwrap         Don't wrap words to the next line automatically
      --verbose            Show timings for response

Environment Variables:
      OLLAMA_HOST                IP Address for the ollama server (default 127.0.0.1:11434)
      OLLAMA_NOHISTORY           Do not preserve readline history
```

PR #5892, opened 2024-07-25
Replacing ibis-framework python package with ibis-framework[pandas]

balajialg:
So, I tested the following [notebook](https://github.com/espm-157/climate-python-template/blob/main/climate.ipynb) from ESPM 157 Github repo in Datahub-staging and realized that the package they need to get the notebook cells to execute successfully is ibis-pandas instead of ibis-framework. I have reached out to Carl via https://github.com/berkeley-dsep-infra/datahub/issues/5827#issuecomment-2251320530 to gain clarity about removing ibis-framework package.

https://pypi.org/project/ibis-framework/


balajialg:
Sigh, need to figure out alternative ways to install ibis-pandas - https://anaconda.org/conda-forge/ibis-pandas?

PR #5945, opened 2024-08-12
[DH-337] upgrade R, R studio to latest version on hubs

shaneknapp:
https://jira-secure.berkeley.edu/browse/DH-337

~fwiw, biology is gonna be a chore...  that image basically needs to be completely redone from scratch.~

bio was actually much easier than anticipated, but boy-howdy that `Dockerfile` is a mess.

### DIFF
--- a/deployments/biology/image/Dockerfile
+++ b/deployments/biology/image/Dockerfile
@@ -1,4 +1,4 @@
-FROM buildpack-deps:focal-scm
+FROM buildpack-deps:jammy-scm
 
 ENV TZ=America/Los_Angeles
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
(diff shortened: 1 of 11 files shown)


ryanlovett (commented):
This looks thorough. We won't be able to anticipate libraries that are incompatible with this change. We should test various R graphics outputs in staging.

Could you please also bump shiny-server from 1.5.21 to 1.5.22 in the shiny Dockerfile? Mostly this is to avoid another CI run, but ideally would should bump shiny-server at the same time as RStudio and R.

ryanlovett:
Actually, since the shiny test already passed, we should delay shiny-server bump to a different PR to prevent another CI run for all affected images.

balajialg (approved):
Thanks, wish I had communicated the R upgrade as part of the onboarding email. We should probably test multiple course notebooks to verify whether RStudio/Shiny works fine in Datahub, R, Public Health and ISchool Hubs. Sharing some course repos for testing,

Datahub/R hub:
https://github.com/ds-modules/PS3-SU23/tree/main
https://github.com/ds-modules/ECON-140-FA22-RDE/tree/main

Ischool Hub:
https://github.com/UCB-MIDS/w241

PH Hub:
https://github.com/ph142-ucb/ph142-su24

Shiny Hub:
https://github.com/ds-modules/polisci-3

shaneknapp:
> Actually, since the shiny test already passed, we should delay shiny-server bump to a different PR to prevent another CI run for all affected images.

tbf, it's not a big deal and i'll cancel the current build(s) and make the change(s).

PR #5983, opened 2024-08-20
[DH-359] bumping seaborn

felder:
https://jira-secure.berkeley.edu/browse/DH-359

In addition to bumping seaborn, I'm also bumping the RAM for admins cuz 2GB was not enough for mamba install/upgrade to succeed and being able to do that for testing is quite useful.


felder:
See issue #5827

PR #5986, opened 2024-08-21
[DH-360] bumping placeholder values

shaneknapp:
https://jira-secure.berkeley.edu/browse/DH-360

@balajialg please review these values...  we might want to bump a couple of these hubs up past 1 but that will be on a hub-by-hub basis.

### DIFF
--- a/node-placeholder/values.yaml
+++ b/node-placeholder/values.yaml
@@ -95,7 +95,7 @@ nodePools:
       requests:
         # Some value slightly lower than allocatable RAM on the nodepool
         memory: 60929654784
-    replicas: 0
+    replicas: 1
   data100:
     nodeSelector:
       hub.jupyter.org/pool-name: data100-pool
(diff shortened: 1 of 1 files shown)


balajialg:
@shaneknapp CE 70 (https://classes.berkeley.edu/content/2024-fall-civeng-70-001-lec-001) will be taught this semester and it will use CEE hub. We can either bump up the placeholder value for this hub or manage it with calendar events (around 100 students). Otherwise, Values look good to me.

shaneknapp:
> @shaneknapp CE 70 (https://classes.berkeley.edu/content/2024-fall-civeng-70-001-lec-001) will be taught this semester and it will use CEE hub. We can either bump up the placeholder value for this hub or manage it with calendar events (around 100 students). Otherwise, Values look good to me.

i'll bump it to 1, and we can adjust this if we need to via calendar event(s).

balajialg:
@shaneknapp sounds like a plan

shaneknapp:
@balajialg btw, i wanted to point out that i set the `dev` hub to `0`...

balajialg:
@shaneknapp thanks for the nudge! It should be good since discovery interns completed their work. I will add a calendar event whenever I need to access the dev hub.

PR #6023, opened 2024-08-26
porting things to niquests

shaneknapp:
i'd like to start here before fixing all of the images that have `requests` installed....

i tested the python in `scripts` and things work exactly as intended.  i also tested the scaler via the steps [here](https://docs.datahub.berkeley.edu/admins/howto/calendar-scaler.html#working-on-testing-and-deploying-the-calendar-scaler), and again, everything is working as intended.

this resolves https://github.com/berkeley-dsep-infra/datahub/security/dependabot/182


ryanlovett (approved):
I think this makes sense for how requests is used in these places. If niquests isn't as compatible as we think, we can replace with python's http.client, or aiohttp.

shaneknapp:
> I think this makes sense for how requests is used in these places. If niquests isn't as compatible as we think, we can replace with python's http.client, or aiohttp.

tbf, it really does seem to be as compatible...  :)

another thing i was thinking of doing was importing niquests directly, and then s/requests/niquests/g in all of the scripts...  but since this work and was a much simpler change, i'm inclined to leave things as they are in this PR.

ryanlovett:
> another thing i was thinking of doing was importing niquests directly, and then s/requests/niquests/g in all of the scripts... but since this work and was a much simpler change, i'm inclined to leave things as they are in this PR.

Yeah, I thought the same thing. "niquists as requests" kind of documents why it is being used.

PR #6030, opened 2024-08-27
[DH-301] hopefully fixing bash arrays + parsing

shaneknapp:
after checking out the results/execution logs of one of the new github workflows (https://github.com/berkeley-dsep-infra/datahub/actions/runs/10584196501/job/29327903196) i noticed that the hub names to deploy weren't properly being added to the array i defined called `HUBS`:

image

this is odd, and i even tested locally and it *should* have properly populated the list:

```
(dh) ➜  workflows git:(more-bash-logic-shennanigans) ✗ a=(); a+="asdf"; a+="fdas"; echo $a
asdf fdas
```

so...  i think the extra space in this expression will force things to work:  `HUB+="$hub "

however, this is a sign that i think we should NOT use bash for stuff like this and instead have some small python helper scripts that will at least behave in an expected way.

### DIFF
--- a/.github/workflows/deploy-hubs.yaml
+++ b/.github/workflows/deploy-hubs.yaml
@@ -37,8 +37,8 @@ jobs:
             # deploy any hubs that have been labeled for deployment
             for label in $(echo -e "${{ steps.pr-labels.outputs.labels }}"); do
               if [[ "$label" == hub-* ]]; then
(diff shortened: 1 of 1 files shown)


shaneknapp:
i manually added two hub labels for no-op testing when i merge to staging.

shaneknapp:
ok that seemed to work!

image

PR #6038, opened 2024-08-28
update version of stat20data package

andrewpbray:


andrewpbray:
@balajialg @ryanlovett @shaneknapp 

This is an urgent request to have a particular data set on the hub (that lives in `stat20data`) accessible by 9 am tomorrow.

Apologies for this appearing so late. I have a difficult time remembering that I need to update two separate docker images whenever I change any dependencies in the course: [the docker image](https://github.com/stat20/stat20-docker) that compiles the course curriculum and the docker image here that the students use.

andrewpbray:
thank you thank you!

PR #6053, opened 2024-08-30
[DH-361] adding nature hub

felder:
https://jira-secure.berkeley.edu/browse/DH-361


shaneknapp (approved):
LGTM+++

looks absolutely complete, and thanks for the doc updates.  i guess we merge, and see what happens next?  ;)

PR #6110, opened 2024-09-06
Use new deployment configuration for shiny

ryanlovett:


.github/workflows/deploy-hubs.yaml
```diff
@@ -103,7 +103,7 @@ jobs:
             echo "Deploying single-user image and hub config to ${deployment}"
             hubploy --verbose deploy --timeout 30m ${deployment} hub staging
             echo
-          done < <(python .github/scripts/determine-hub-deployments.py --only-deploy logodev stat159 stat20 nature a11y ugr01 data101 astro biology cee)
+          done < <(python .github/scripts/determine-hub-deployments.py --only-deploy logodev shiny stat159 stat20 nature a11y ugr01 data101 astro biology cee)
```
shaneknapp: put this in a 2nd PR, we don't want to accidentally deploy before the image is built

shaneknapp: you also need to add this to the second job to ensure that it's deployed to prod (again in the 2nd PR)

.circleci/config.yml
shaneknapp, line 1: also delete the gradebook entries in circleci

shaneknapp (changes requested):

Issue #6121, opened 2024-09-09
Request more RAM for Stat 131a


ryanlovett:
**Which hub do you want more RAM on? Please link the URL to the hub while specifying the hub name**

r.datahub.berkeley.edu

**Which class is this request for?**

2024 Fall, Stat 131a
**What is the Academic Guide URL for this class?**

https://classes.berkeley.edu/content/2024-fall-stat-c131a-001-lec-001

**What is the bCourses URL for this class?**

https://bcourses.berkeley.edu/courses/1537932

**How many students do you expect in this class?**

An approximate number would do.

86

**How much RAM per user is needed?**

2GB

**Why does this class need this much RAM?**

Knitting larger .Rmd files.

**By what date (MM/DD) do you want this request to be fulfilled?**

ASAP

**Until when (MM/DD) do you require the additional RAM?**

End of the term, 06/01

**Any additional information we should know about?**

Related thread in `#ucb-datahub`

balajialg:
@ryanlovett Closing this issue as the changes are merged to prod.

Issue #6094, opened 2024-09-05
Request more RAM for Demog 213


wrathofquan:
**Which hub do you want more RAM on? Please link the URL to the hub while specifying the hub name**

For example: Datahub (datahub.berkeley,edu)

**Which class is this request for?**

[Demog 213](https://github.com/berkeley-demography/demog-213-f24)

**What is the Academic Guide URL for this class?**

[https://classes.berkeley.edu/content/2024-fall-demog-213-001-dem-001](https://classes.berkeley.edu/content/2024-fall-demog-213-001-dem-001)

**What is the bCourses URL for this class?**

[https://bcourses.berkeley.edu/courses/1539545/](https://bcourses.berkeley.edu/courses/1539545/)

**How many students do you expect in this class?**

10

**How much RAM per user is needed?**

4GB of RAM per student.

**Why does this class need this much RAM?**

We will be working with larger datasets for students to practice R

**By what date (MM/DD) do you want this request to be fulfilled?**

Ideally by 9/10/2024

**Until when (MM/DD) do you require the additional RAM?**

12/20/2024

**Any additional information we should know about?**

balajialg:
@wrathofquan PR to increase RAM for Demog 214 got merged to staging hub - you can test the changes in https://staging.datahub.berkeley.edu/

wrathofquan:
Thanks @balajialg. For some reason this staging hub link is not pulling from the course repo (https://github.com/berkeley-demography/demog-213-f24). Is that expected?

Also, I am showing 8GB of RAM which might be a holdover from a previous semester's request. 
<img width="540" alt="image" src="https://github.com/user-attachments/assets/402e0b77-b68a-4b87-a736-67e72b9a3fef">

balajialg:
@wrathofquan That's weird! I can pull the R script to R Hub Staging using the following [nbgitpuller link](https://r-staging.datahub.berkeley.edu/hub/user-redirect/git-pull?repo=https%3A%2F%2Fgithub.com%2Fberkeley-demography%2Fdemog-213-f24&branch=main&urlpath=rstudio%2F). Can you check whether this link works for you? If not, what is the error message you are getting?

[...]

Issue #6015, opened 2024-08-26
Request more RAM for class EE 120


edwardneo:
**Which hub do you want more RAM on? Please link the URL to the hub while specifying the hub name**

EECS hub (eecs.datahub.berkeley.edu)

**Which class is this request for?**

2024 Fall, EE 120

**What is the Academic Guide URL for this class?**

https://classes.berkeley.edu/content/2024-fall-eleng-120-001-lec-001

**What is the bCourses URL for this class?**

https://bcourses.berkeley.edu/courses/1537177

**How many students do you expect in this class?**

170

**How much RAM per user is needed?**

4GB of RAM per student.

**Why does this class need this much RAM?**

We have signal processing labs that involve working with large amounts of data.

**By what date (MM/DD) do you want this request to be fulfilled?**

09/13

**Until when (MM/DD) do you require the additional RAM?**

12/20

**Any additional information we should know about?**

Nope!

balajialg:
@edwardneo Thanks, I will get this request done during the week of 9/9 - 9/13

balajialg:
@edwardneo PR to increase RAM for Eleng 120 got merged to staging hub - you can test the changes in https://eecs-staging.datahub.berkeley.edu/

balajialg:
Closing this issue as the changes are merged to prod. Please reopen if there are any issues.

PR #6123, opened 2024-09-09
Increase RAM for Demog 213, Data C131A and Eleng 120

balajialg:
Resolves 
https://github.com/berkeley-dsep-infra/datahub/issues/6121
https://github.com/berkeley-dsep-infra/datahub/issues/6094
https://github.com/berkeley-dsep-infra/datahub/issues/6015


ryanlovett (changes requested):
I think the Data 131a stanza needs to go into R hub.

shaneknapp (approved):

PR #6169, opened 2024-09-12
[DH-3] bumping chp ram again

shaneknapp:
https://github.com/jupyterhub/configurable-http-proxy/issues/388#issuecomment-2346951642


felder (changes requested):
suggest we set guarantee = limit = 3g and make this change in individual hub configs instead of globally.

shaneknapp:
> suggest we set guarantee = limit = 3g and make this change in individual hub configs instead of globally.

done

felder (approved):

PR #6192, opened 2024-09-17
[DH-3] Set chp timeout for all hubs to 5 sec

shaneknapp:
after chatting w/ @consideRatio and @minrk, @felder and i decided to try setting some low timeouts.

we've manually deployed the config changes to datahub and data8 prod, and are monitoring the number of open ports in the `chp`.

things look...  good for now.  we've gone from ~1X,XXX open connections to the hub pod to ~only a couple of hundred a few thousand (roughly 6k)~ some number between ~6k and ~60:

~180 users currently:
```
/srv/configurable-http-proxy $ netstat -natp | grep <hub pod public ip> | wc -l
5918
```

~250 users currently:
```
/srv/configurable-http-proxy $ netstat -natp | grep <hub pod public ip> | wc -l
65
```

this number is holding steady with ~250 users on the impacted hubs.

we'll watch this today, esp during peak times and see if this helps keep the number of ports under control.

### DIFF
--- a/deployments/data100/config/common.yaml
+++ b/deployments/data100/config/common.yaml
@@ -10,6 +10,9 @@ jupyterhub:
         hub.jupyter.org/pool-name: core-pool-2024-05-08
   proxy:
     chp:
+      extraCommandLineFlags:
+        - "--timeout=5000"
+        - "--proxy-timeout=5000"
       nodeSelector:
         hub.jupyter.org/pool-name: core-pool-2024-05-08
       resources:
--- a/deployments/data100/config/common.yaml
+++ b/deployments/data100/config/common.yaml
@@ -10,9 +10,6 @@ jupyterhub:
         hub.jupyter.org/pool-name: core-pool-2024-05-08
   proxy:
     chp:
-      extraCommandLineFlags:
-        - "--timeout=5000"
-        - "--proxy-timeout=5000"
       nodeSelector:
         hub.jupyter.org/pool-name: core-pool-2024-05-08
       resources:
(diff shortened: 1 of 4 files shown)


shaneknapp:
we're going to need to wait until later in the day to see if this has a positive impact for our users...

shaneknapp:
after 90m, data8 has ~300 users and is showing just under 24k ports open to the public ip.  the port count is slowly rising, and the user count is steady:

```
/srv/configurable-http-proxy $ netstat -natp | grep <hub pod public ip> | wc -l
23685
```

shaneknapp:
> after 90m, data8 has ~300 users and is showing just under 24k ports open to the public ip. the port count is slowly rising, and the user count is steady:
> 
> ```
> /srv/configurable-http-proxy $ netstat -natp | grep <hub pod public ip> | wc -l
> 23685
> ```

after a bit, it drops down to ~16k.  after another bit, back up to 22-23k...  then back down to ~16k.

HERE BE DRAGONS

shaneknapp:
anyways...  this change seems to help?  we will definitely be tracking this over the course of the next couple of week.

however, it seems benign enough that i will most likely make this change across all hubs and push to prod ~EOD today.

shaneknapp:
> > after 90m, data8 has ~300 users and is showing just under 24k ports open to the public ip. the port count is slowly rising, and the user count is steady:
> > ```
> > /srv/configurable-http-proxy $ netstat -natp | grep <hub pod public ip> | wc -l
> > 23685
> > ```
> 
> after a bit, it drops down to ~16k. after another bit, back up to 22-23k... then back down to ~16k.
> 
> HERE BE DRAGONS

i am continuing to see this exact behavior on all three of these hubs.  drop to ~12k-18k, growth to ~23k over the course of an hour, back down to 12k-18k.

PR #6222, opened 2024-09-19
[DH-377] bumping jupyter core packages, moving some from pip to conda

shaneknapp:
since we're not seeing the buildup of ephemeral ports on hubs w/newer versions of jupyterlab, etc, we'll try this on one of our impacted hubs


felder:
@shaneknapp should we bump jupytext to 1.16.4, also did you try a local build first?

PR #6226, opened 2024-09-20
remove myself from the issue template for archive file requests

shaneknapp:
since i don't perform these tasks, i'd like to cut down on the email and slack spam a bit.

@balajialg you ok w/this?


balajialg:
No worries!

PR #6244, opened 2024-09-24
[DH-377] bump data101 proxy ram

shaneknapp:
since they had an outage, mimic the same parameters we use w/datahub, data100 and data8.

we should also consider setting this to 3G for all hubs.


felder (approved):
LGTM

PR #6246, opened 2024-09-24
[DH-301] preparing to move edx to its own image repo

shaneknapp:
this change will also move the image from gcr to artifact repository...


shaneknapp:
i created a new secret for hubploy and stored it in the org as "GAR_SECRET_KEY_EDX", and enabled artifact registry in the data8x-scratch project.

PR #6268, opened 2024-09-25
[DH-301] new workflow for edx deployment

shaneknapp:
this is in place of https://github.com/berkeley-dsep-infra/datahub/pull/6256

i also deleted the circleci webhook, so we shouldn't get any more errors.


shaneknapp:
i figured it out...  hubploy handles its own authentication, so instead of using the `google-github-actions/auth@v2` and associated secrets, we just let it do it's own thing.  i'll fix this up and we'll be g2g.

Issue #6316, opened 2024-10-01
Data 100 Grad Project Dataset


malavikhasudarshan:
### Your Name

Malavikha Sudarshan

### Affiliated Course Name

Data C100

### bCourses ID

1537664

### Role

Lead Infrastructure Teaching Assistant

### Hub URL

datahub.berkeley.edu

### Shared Directory Name

data-100-readwrite, data-100-shared

### End Date

12/20/24

### Request Deadline

ideally 10/01, (this was only decided in the early hours of this morning by our staff so this is quite last-min; if it will take longer i can let the content team know)

balajialg:
@malavikhasudarshan Changes are merged to prod now and will be live in few mins. Please test whether you have access to `data100-shared` and `data100-shared-readwrite` directories in your home dir (Please log out and log back in)

balajialg:
Closing this issue as the changes are merged to prod

PR #6002, opened 2024-08-23
Add shared-read-write directories in DataHub for COMPSS 214A course

balajialg:
Ref: https://jira-secure.berkeley.edu/browse/DH-311

Created the shared directories for this course (in staging and prod) in NFS server

PR #6317, opened 2024-10-01
Enable shared directories for data100 based on bcourses id

balajialg:
Even though Data 100 already has a shared directory, I thought it would be good to enable the bcourses-based workflow for creating shared directories. Students who are no longer affiliated with the course at the end of the semester won't have access to the shared directory.

Created `data100-shared-readwrite` and `data100-shared` directories in NFS server

Ref: https://github.com/berkeley-dsep-infra/datahub/issues/6316

### DIFF
--- a/deployments/data100/config/common.yaml
+++ b/deployments/data100/config/common.yaml
@@ -116,18 +116,18 @@ jupyterhub:
       course::1537664::enrollment_type::teacher:
         extraVolumeMounts:
           - name: home
-            mountPath: /home/jovyan/data-100-readwrite
-            subPath: _shared/course/data-100
+            mountPath: /home/jovyan/data100-shared-readwrite
+            subPath: _shared/course/data100-shared-readwrite
       course::1537664::enrollment_type::ta:
         extraVolumeMounts:
           - name: homes
-            mountPath: /home/jovyan/data-100-readwrite
-            subPath: _shared/course/data-100
+            mountPath: /home/jovyan/data100-shared-readwrite
+            subPath: _shared/course/data100-shared-readwrite
       course::1537664::enrollment_type::student:
         extraVolumeMounts:
           - name: home
-            mountPath: /home/jovyan/data-100-shared
-            subPath: _shared/course/data-100
+            mountPath: /home/jovyan/data100-shared
+            subPath: _shared/course/data100-shared
             readOnly: true
     admin:
       mem_guarantee: 2G
--- a/deployments/data100/config/common.yaml
+++ b/deployments/data100/config/common.yaml
@@ -111,6 +111,24 @@ jupyterhub:
       course::1524699::group::all-admins:
         admin: true
         mem_limit: 4G
+
+      # Data C100, Fall 2024, https://github.com/berkeley-dsep-infra/datahub/issues/6316
+      course::1537664::enrollment_type::teacher:
+        extraVolumeMounts:
+          - name: home
+            mountPath: /home/jovyan/data-100-readwrite
+            subPath: _shared/course/data-100
+      course::1537664::enrollment_type::ta:
+        extraVolumeMounts:
+          - name: homes
+            mountPath: /home/jovyan/data-100-readwrite
+            subPath: _shared/course/data-100
+      course::1537664::enrollment_type::student:
+        extraVolumeMounts:
+          - name: home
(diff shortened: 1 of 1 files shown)


ryanlovett:
@balajialg I think you can still use the existing path as on the other hubs, `shared-readwrite` and `shared`, but you can move it from `admin` into the bcourse stanza. By reusing `shared-readwrite` and `shared`, we won't have to name it differently in the docs.

If you did want to rename it though, it should be `data100` and not `data-100`.

balajialg:
@ryanlovett I went by the TA's inputs for naming the shared directory - https://github.com/berkeley-dsep-infra/datahub/issues/6316. They preferred data-100 as part of the github issue.

Also, Data 100 Hub is used by two courses - Data 100 and Econ 148. It would make sense to have a shared directory that is specific to Data100 right (instead of the `shared-readwrite` and `shared` which is shared between two courses)

ryanlovett:
Okay, thanks for that context. I didn't know that Econ 148 used data100.

It makes sense to have a shared directory that is specific to data100, but it could still be mounted as `shared` and `shared-readwrite`. The directory on the NFS server could be named whatever it needs to be.

balajialg:
@ryanlovett Thanks, Any reason for preferring to name `shared` over `data100-shared`? I can honestly go with either of the options but would prefer to name it `data100-shared`.

felder:
@balajialg the naming matters very little and only comes into play if there are multiple shared dirs in a hub with different people needing to access each.

IMO the naming really doesn't matter a whole lot from an operational perspective.  If the customers want data-100-shared vs data100-shared or shared or whatever, I don't really see that it makes a big difference.

If you want a specific guideline, I'd go with COURSENAME-shared-readonly/COURSENAME-shared-readwrite, but again the convention is really not a huge deal.  It's pretty obvious from the configs what's happening.

balajialg:
@felder Yes, I am also inclined to go with the naming that the user wants and call it a day. 

I am asking follow up questions to understand @ryanlovett's rationale for the suggestions proposed.

felder:
@balajialg his rationale covers two points here.  First, he was assuming that for data100 since it has a dedicated hub that there was only a single course (turns out that is not the case) which makes putting the course in the name redundant.  Second whenever we refer to data100 everywhere...it's data100 not data-100.

balajialg:
@felder Cool, thanks for chipping in. I have named it `data100-shared` as per his previous wish. I think this PR is g2g unless there is more details to be hashed out.

felder:
@balajialg it does occur to me that one thing that this might impact is the archiver which is supposed to skip shared directories.  Consistently naming them will allow us to filter them out with greater accuracy.  Probably safe to just assume that shared directories will contain the word "shared" in them which can be used for an exclusion, but if a specific convention is required my vote would be to do COURSENAME-shared-readonly/COURSENAME-shared-readwrite

Such a convention might also help customers who happen to have involvement with a hub that contains multiple shared directories for multiple courses (an uncommon occurrence, but a foreseeable one)

ryanlovett:
One reason for just using `shared` instead of `COURSENAME-shared` is that that is how we've named them thus far across all hubs, and it is easier to document how course staff should use such directories. But if we were to adopt a new convention, then `COURSENAME-shared` does make sense too, especially for shared hubs, and to fix the convention for `COURSENAME` to be `deptNNN` since that is how we name the hubs. I suppose the latter could be `dept-NNN` (tomato/tomahto, yml/yaml), but it'd be good to have one naming scheme.

balajialg:
@ryanlovett @felder Fair enough, Lets fix `COURSENAME-shared` and `COURSENAME-shared-readwrite` as a convention for adding shared directories to address future requests. I will fix the template to reflect the same. Can this PR get merged to staging considering the time sensitive nature of the request?

felder:
I'm curious how this PR will behave considering it's using the same mechanism that didn't seem to work for whatever reason with resource allocation for data100.

balajialg:
@felder Thanks, It seems to have been working fine for our previous requests to add shared directories so fingers crossed.

Eg: https://github.com/berkeley-dsep-infra/datahub/pull/6257
https://github.com/berkeley-dsep-infra/datahub/pull/6206
https://github.com/berkeley-dsep-infra/datahub/pull/6002
https://github.com/berkeley-dsep-infra/datahub/pull/5896

PR #6329, opened 2024-10-02
removing extra CNAME

shaneknapp:


ryanlovett (approved):
I think the top-level file is not needed. GitHub Pages publication definitely did not work with the top-level file and without `docs/CNAME`.

PR #6347, opened 2024-10-08
[DH-227] in preparation for running git precommits

shaneknapp:
i installed `ruff`, which is a linter and python formatter.  this is the result of `ruff format`


shaneknapp:
i'm just going to merge this...  my test runs of the pre-commit hooks are going to make a bunch of additional changes, which we'll need to be reviewed separately.

Issue #6349, opened 2024-10-09
Request update to R package patchwork for class ph142


paulamarm:
### Package Name

patchwork

### Hub URL

publichealth.datahub.berkeley.edu

### Course Name

Introduction to Probability and Statistics in Biology and Public Health

### Semester Details

Ongoing: Just needs update

### Installation Deadline

Ideally as soon as possible to prevent the instructors from manually making the update every time they use it in materials for lecture!
Currently seeing the following errors appear:
```{r first-sample-250-error, echo=F, eval=F}
## If you are getting an error message like the following:
# Error in Ops.data.frame(guide_loc, panel_loc) :
#'==' only defined for equally-sized data frames
## ... then the issue might be a stale version of the 'patchwork' library. 
## You may be able to clear it up by manually running the following *once*, and then restarting R:
remove.packages('patchwork')
install.packages('patchwork')
```
Thank you!

felder:
@paulamarm can you provide me with what I'd need to enter into the R console to produce your error?

felder:
@paulamarm also can you please confirm the version of patchwork you're requesting?

balajialg:
@paulamarm We figured out that cloning the [PH142-fa24 github repository](https://github.com/ph142-ucb/ph142-fa24) and launching the [following notebook](https://github.com/ph142-ucb/ph142-fa24/blob/efbd1b21bf12ed280978bbf86880ae0ac0c9a15c/lec/Day-11_Intro-probability/Ch09_Intro-probability.Rmd#L257) allowed us to reproduce the issue that you reported. We would appreciate sharing such test cases as part of the request since it takes a lot of admin time to reproduce the issue with minimal information provided. In addition, we were unclear about the version that the package needs to get upgraded to (we eventually figured that it was version 1.3.0)

shaneknapp:
built, tested and about to merge to prod:
```
> packageVersion('patchwork')
[1] ‘1.3.0’
```

https://github.com/berkeley-dsep-infra/datahub/pull/6356

felder:
@paulamarm ok I actually made some significant changes to the image in order to fix a few other R packages that I noticed didn't install successfully as well as to update a number of components.

Details, if you're curious:  
https://github.com/berkeley-dsep-infra/publichealth-user-image/pull/5

[...]

PR #6362, opened 2024-10-10
[DH-394] merge #6361 to prod

felder:
https://jira-secure.berkeley.edu/browse/DH-394 #6349


felder:
Seems to work ok.  Not sure how much time we want to provide for independent testing from the reporter.  Other than that, I think this good to go.

shaneknapp (approved):

shaneknapp:
lgtm, if we don't hear back on #6349 we can merge to prod EOW

PR #6382, opened 2024-10-17
Disable jupyter lab scrollPastEnd.

ryanlovett:
This may prevent buggy scrolling behavior on some notebooks.


balajialg:
@ryanlovett Wow, good stuff reproducing that issue. I couldn't reproduce it when I tried at my end.

ryanlovett:
@balajialg It was happening to me when I was doing something completely different, so I was luckily/unluckily able to use it in this case.

balajialg:
That's awesome @ryanlovett

shaneknapp:
same, i couldn't reproduce it either!

PR #6386, opened 2024-10-18
[DH-227] apply changes from running `pre-commit run --all-files`

shaneknapp:
<3 to massive PRs


ryanlovett (approved):
I don't see anything obviously wrong in the code changes. I think it can be merged as is.

Another way would be to separate the changes into different PRs, e.g. one for line endings, one for imports, etc. That would just be to make things easier to review. But again, it seems fine.

shaneknapp:
> Another way would be to separate the changes into different PRs, e.g. one for line endings, one for imports, etc. That would just be to make things easier to review. But again, it seems fine.

yeah, i agree but given how much time i've already dedicated to this i'm down to just do it this way...  :)

balajialg (approved):
doc changes LGTM!

felder (changes requested):
version lock in requirements.txt and dev-requirements.txt?  Also scripts/infra-packages/requirements.txt has some interesting re-ordering, but we can probably just get rid of it.

shaneknapp:
@felder feel free to merge if you're happy w/the changes to `requirements.txt` and `dev-requirements.txt`

PR #6424, opened 2024-10-25
Update CI/CD doc

ryanlovett:
~~Primarily, add the doc to the navigation.~~ Then use markdown features where it makes sense. (terms, callout)

### DIFF
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -35,6 +35,7 @@ website:
           - admins/storage.qmd
           - admins/cluster-config.qmd
           - admins/credentials.qmd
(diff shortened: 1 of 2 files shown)


ryanlovett:
Looks like you added it to the nav while I was making this PR!

shaneknapp (commented):
looks great, thanks!  could you possibly add a little extra detail about quarto workflow and how that fits in to our CI/CD?

if you don't think it needs it, or it's described in detail elsewhere, that's fine.

ryanlovett:
@shaneknapp Good idea, yes. I'll add something...

PR #6436, opened 2024-10-28
clean up core pool sysctl config

shaneknapp:
fix the formatting of the core pool sysctl config, delete an accidental check-in and explicitly git ignore it.


felder:
Do pods inherit these settings?  Also I don't think it's a good idea to do this at the node level.

felder (commented):
Do pods inherit these settings? Also I don't think it's a good idea to do this at the node level. I think it'd be better to have a deployment task that modifies the chp deployment. Even if pods do inherit, we don't want all pods on the core pool to do this. Nor would we want to expand the ephemeral port range on the node itself.

shaneknapp:
> Do pods inherit these settings? Also I don't think it's a good idea to do this at the node level. I think it'd be better to have a deployment task that modifies the chp deployment. Even if pods do inherit, we don't want all pods on the core pool to do this. Nor would we want to expand the ephemeral port range on the node itself.

these are applied to the node, and then inherited by the pods that run on said nodes.

your points are valid though, we probably don't want both all the pods on these nodes (hub, k8s/gke plumbing) to inherit the enlarged ephemeral port range.

it looks like the only way to do this is in `hub/values.yaml`, in `proxy:` and setting [chp.extraPodSpec](https://z2jh.jupyter.org/en/latest/resources/reference.html#proxy-chp-extrapodspec) to use this value and then redeploying the hubs.  source docs are terrible, but the extrapodspec docs are here (search for `securityContext.sysctls`): https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#PodSpec

fwiw, we will leave the remaining sysctl settings in the core pool config -- that is meant to tune the nodes themselves.

shaneknapp:
i'll revert the addition of the port range, and update the title/desc of this PR and merge it.  i'd like to preserve the formatting fix and deletion of the old file that was accidentally checked in.

PR #6441, opened 2024-10-30
[DH-377] Increase ephemeral ports by default

shaneknapp:
instead of manually patching proxy pods that run out of ephemeral ports in a terrible game of whack-a-mole, just give every hub's chp more ports by default.

it's a band-aid for https://github.com/jupyterhub/configurable-http-proxy/issues/557, and while it doesn't fix the problem outright it definitely lessens the impact significantly.

instead of ~15-45m of "503 service unavailable" being presented to the users when we hit the default limit of 28231 ephemeral ports (60999 - 32768), we give ourselves 55000 ephemeral ports (65000 - 10000).  this limits downtime to a few minutes of high hub response times, and possibly a few users seeing 503.

we haven't seen any impact on other services, and remain slightly optimistic of it's efficacy in smoothing out the road.

we've also given the worst impacted chp pods 3G of ram each, but could also consider scaling that down to 2.5G.  under max load, when the port count is ~32000, the chp pod ram usage is typically between ~1.2-1.6G.

i will test this on logodev.d.b.e

### DIFF
--- a/hub/Chart.yaml
+++ b/hub/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: '1.0'
 description: Deployment Chart for JupyterHub
 name: hub
-version: 20240731-224556.git.8734.h9b451d16
+version: 20240731-224556.git.8735.h3e417283
--- a/hub/Chart.yaml
+++ b/hub/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: '1.0'
 description: Deployment Chart for JupyterHub
 name: hub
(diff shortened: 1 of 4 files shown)


shaneknapp:
deployed to logodev.d.b.e and the updated port range has been applied!

```
/srv/configurable-http-proxy $ sysctl -a|grep range
net.ipv4.ip_local_port_range = 10000    65000
```

felder (approved):
LGTM

shaneknapp:
this took down all hubs that already had the `kubectl patch` in place.

other hubs were unaffected.

ofc the hubs that went down are the ones that get the most usage.  :)

Issue #6498, opened 2024-11-19
Request PyTorch for Data 100


malavikhasudarshan:
### Package Name

pytorch

### Package Version

2.5.1

### Hub URL

data100.datahub.berkeley.edu

### Reproducible test case

import torch

print(torch.__version__)

### Course Name

https://classes.berkeley.edu/content/2024-fall-data-c100-001-lec-001

### Semester Details

Fall 2024

### Installation Deadline

(12/02)

shaneknapp:
this is now live on data100-staging

shaneknapp:
pushed to prod

PR #6499, opened 2024-11-19
update data100 image tag to 6ffa8f224124

shaneknapp:
update data100 image tag to 6ffa8f224124

deployments/data100/hubploy.yaml


shaneknapp:
resolves https://github.com/berkeley-dsep-infra/datahub/issues/6498

PR #6536, opened 2024-12-13
Update gradebook image

andrewpbray:
Hey @ryanlovett , I got the sha from here:

https://hub.docker.com/r/gradebook/gradebook/tags

Think this was our revised workflow when we ditched `latest`?


andrewpbray:
@ryanlovett If this can be merged within the next 10 min, we can use it at the Gradebook lunch at noon. If not, that's ok, we can just describe the new functionality and demo locally.

ryanlovett:
Sorry @andrewpbray , and thanks @balajialg ! Just seeing this now.

andrewpbray:
Thanks!

PR #6549, opened 2024-12-18
[DH-420] Remove bcourses based configs across hubs

balajialg:
WIP PR - To be merged during maintenance window


deployments/datahub/config/common.yaml
```diff
@@ -229,57 +135,6 @@ jupyterhub:
       # DataHub Infrastructure staff
       # https://bcourses.berkeley.edu/courses/1524699/groups#tab-80607
       course::1524699::group::all-admins:
+        admin: true
```
ryanlovett, line 138: We should use discrete scopes instead of setting admin like this.

ryanlovett (approved):
This looks good.

IMO we should not use admin=true. We should use the datahub infrastructure group to grant explicit administrative scopes.

deployments/datahub/config/common.yaml
balajialg: @ryanlovett Thanks for highlighting this change. We still have `admin:true` flag enabled in data8, data100, data 101, publichealth, eecs, stat 20, stat 159, dev and a11y hubs. I will remove `admin:true` flag across all configs in my upcoming PR.

ryanlovett: @balajialg I feel like I periodically remove it and then it creeps back in.

PR #6588, opened 2025-01-08
[DH-435] fixing empty stanza and adding datahub staff as admins

shaneknapp:
i tested this by manually deploying to astro-staging.datahub.berkeley.edu

### DIFF
--- a/deployments/astro/config/common.yaml
+++ b/deployments/astro/config/common.yaml
@@ -91,6 +91,12 @@ jupyterhub:
   # formerly jupyterhub.hub.extraConfigMap
   custom:
     group_profiles:
+      # DataHub Infrastructure staff
+      # https://bcourses.berkeley.edu/courses/1524699/groups#tab-80607
+      course::1524699::group::all-admins:
+        admin: true
+        mem_limit: 4G
+
     admin:
       extraVolumeMounts:
         - name: home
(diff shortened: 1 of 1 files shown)


deployments/astro/config/common.yaml
```diff
@@ -91,6 +91,12 @@ jupyterhub:
   # formerly jupyterhub.hub.extraConfigMap
   custom:
     group_profiles:
+      # DataHub Infrastructure staff
+      # https://bcourses.berkeley.edu/courses/1524699/groups#tab-80607
+      course::1524699::group::all-admins:
+        admin: true
```
ryanlovett: We shouldn't use `admin: true`. We should use the discrete roles in `loadRoles`.

ryanlovett (changes requested):

deployments/astro/config/common.yaml
shaneknapp: ok, i just grabbed this snipped from the data100 config.

in fact, we do this in a bunch of hubs (which i guess we should fix):

```
(dh) ➜  deployments git:(fix-astro-config) ✗ grep -lr "course::1524699::group::all-admins:" *
a11y/config/common.yaml
astro/config/common.yaml
biology/config/common.yaml
data100/config/common.yaml
data101/config/common.yaml
data8/config/common.yaml
datahub/config/common.yaml
dev/config/common.yaml
eecs/config/common.yaml
publichealth/config/common.yaml
r/config/common.yaml
stat159/config/common.yaml
stat20/config/common.yaml
```

Issue #6643, opened 2025-01-15
Shared directory for PS 191


ekhartman:
### Your Name

Erin Hartman

### Affiliated Course Name

PS 191

### bCourses ID

https://bcourses.berkeley.edu/courses/1541486

### Role

Professor

### Hub URL

r.datahub.berkeley.edu

### End Date

- 05/16/2025

### Request Deadline

I'd need this set up before classes start on Tuesday 01/15.

As a note, too, I've listed it as r.datahub, but there were some issues getting the course set up on there (it is temporarily on a different data hub).

balajialg:
@ekhartman Changes are merged to https://r.datahub.berkeley.edu/ and should be live in the next 40 minutes. Any running servers need to be restarted for the change to take effect. Please reopen this issue if you have any issues.

PR #6646, opened 2025-01-16
Enable shared directories for PS 191

balajialg:
https://github.com/berkeley-dsep-infra/datahub/issues/6643

@felder @ryanlovett As discussed during our meeting yesterday, I have set the directory name as bcourses id of the course.


felder (commented):
When we chatted yesterday, we floated the idea of mounting these outside of the user home directory.  Not sure if we should do that here or not.  @ryanlovett thoughts?

Issue #6653, opened 2025-01-16
Giving Data 8 GSIs admin/server access


ella-deguzman:
### Your Name

Ella Deguzman

### Affiliated Course Name

Data 8 

### bCourses ID

1542000

### Role

Content Lead

### Hub URL

datahub.berkeley.edu

### End Date

June 2025 
 @balajialg removed the provided email ids for privacy reasons

### Request Deadline

asap

balajialg:
@ella-deguzman  Changes are merged to https://data8.datahub.berkeley.edu/ and should be live in the next 40 minutes. Any running servers need to be restarted for the change to take effect. Please reopen this issue if you have any issues.

PR #6662, opened 2025-01-17
Enable elevated privileges for data 8 staff

balajialg:
Ref: https://github.com/berkeley-dsep-infra/datahub/issues/6653


balajialg:
@ryanlovett gave a thumps up during a git-credential-helper debug meeting!

Issue #6722, opened 2025-01-28
Elevated Privileges for Data 101 TAs on Data101 DataHub


Zackoon:
### Your Name

Zackary Oon

### Affiliated Course Name

Data 101: Data Engineering

### bCourses ID

1542742

### Role

Head TA

### Hub URL

data101.datahub.berkeley.edu

### End Date

05/16/25

### Request Deadline

01/29

balajialg:
@Zackoon Changes are merged to https://data101-staging.datahub.berkeley.edu/. Please test the changes by logging out of the hub and then logging back in. Let us know if there are any issues.

balajialg:
@Zackoon Changes are merged to https://data101.datahub.berkeley.edu/.

PR #6724, opened 2025-01-28
Add configs to enable elevated privileges for data 101 course staff

balajialg:
Ref: https://github.com/berkeley-dsep-infra/datahub/issues/6722


felder (approved):
looks ok to me

PR #6756, opened 2025-02-06
Dh 410

yijunge-ucb:


docs/tasks/delete-hub.qmd
```diff
@@ -34,18 +34,24 @@ gcloud filestore backups create <hubname>-backup-YYYY-MM-DD --file-share=shares
...
+```
+python3 create_alerts.py --disable_alerts --namespaces <namespace>
+
+```
+9.  Delete k8s namespace:
 ```bash
 kubectl delete namespace <hubname>-staging <hubname>-prod
 ```
-9.  Delete k8s node pool:
+10.  Delete k8s node pool:
```
ryanlovett, line 47: Just an FYI that it is easier to number everything `1.` in the markdown and quarto will do the right thing when rendering. It makes it easier to read diffs for changes, and you don't have to worry about numbering things correctly. No need to change this PR, just letting you know of another way.

ryanlovett (approved):
Looks good, and thanks for documenting! Just one minor comment about numbered lists.

docs/tasks/delete-hub.qmd
yijunge-ucb: Thanks!

PR #6757, opened 2025-02-06
[DH-25] Enable config to test RTC in dev hub

balajialg:
https://jira-secure.berkeley.edu/browse/DH-25

### DIFF
--- a/deployments/dev/config/common.yaml
+++ b/deployments/dev/config/common.yaml
@@ -26,6 +26,11 @@ jupyterhub:
       20-hubHome: |
         # Direct users to named server list
         c.JupyterHub.default_url = '/hub/home'
+        # Enable RTC in Dev hub for testing
+      collaborative: |
+        c.ServerApp.jpserver_extensions = {
+            'jupyter_collaboration': True
+        }
 
     loadRoles:
       # datahub staff


felder (commented):
Would it make sense to put this in staging.yaml until we've tested it?  This is in devhub so it may not matter I guess.  That's really my only concern cuz I'm gonna assume the config is right. ;-)  So if the answer is "no" to staging.yaml, then I approve!

ryanlovett:
Is this configuration needed? i.e. most server extensions are now enabled when you install them.

We used to enable the [TempFileYStore](https://jupyterlab-realtime-collaboration.readthedocs.io/en/latest/configuration.html
) so that the associated sqlite db files were put in /tmp/ rather than user's home directories.

I honestly don't know if we still want this. There can periodically be file locking issues with sqlite files on NFS, but it isn't always the case. It might be worth not adding this back, just to test. IIRC, this /tmp configuration also caused reliability issues that weren't present when it wasn't enabled.

balajialg:
@felder Sure, I will keep this in mind for my future PRs. @ryanlovett Good to know! I have seen this config added during the initial iteration of RTC but wasn't sure about the latest changes. I will close this PR for now.

PS: I remember the crazy times when NFS/Sqlite combo were causing reliability issues.

ryanlovett:
@balajialg It may still be needed. You might try testing with just the package installed.

balajialg:
@ryanlovett Yeah, I did some testing in the dev-staging hub without the config changes and with the package installed. Despite [adding](https://github.com/berkeley-dsep-infra/dev-primary-image/blob/300bc7d90e25dc84187903e75db50f5a5b611161/environment.yml#L17) `Jupyter-Collaboration` package in Dev Hub's primary image, RTC is still not enabled. However, that wasn't the case when I built the image locally on my device.

I will probably hubploy this config change to dev-staging and evaluate whether it works. If it does, I will re-open the PR.

balajialg:
@ryanlovett I just observed that all three profiles in Dev Hub use dev-secondary images? Seems like dev-primary-image is not used by any hubs (ironically). I thought the default was dev-primary-image based on the name.

ryanlovett:
@balajialg It used to be there. Maybe things changed when the image repos were split out? Not sure.

PR #6792, opened 2025-02-13
remove shane from github issue autoassign, add yijun

shaneknapp:


balajialg (approved):
Yes, that makes sense. Thanks @shaneknapp

Issue #6806, opened 2025-02-17
Data 100 Grad Project Dataset


malavikhasudarshan:
### Your Name

Malavikha Sudarshan

### Affiliated Course Name

Data C100

### bCourses ID

1541798

### Role

Lead Infrastructure Teaching Assistant

### Hub URL

data100.datahub.berkeley.edu

### End Date

05/17/25

### Request Deadline

02/18 please! if it will take longer, please let me know and i will ask the professors + graduate TAs to delay the release of this project

balajialg:
@malavikhasudarshan Closing this issue. Please reopen if any inputs are needed!

PR #6811, opened 2025-02-18
[DH-467] Enable shared directory in data100 hub

balajialg:
Addresses https://github.com/berkeley-dsep-infra/datahub/issues/6806

### DIFF
--- a/deployments/data100/config/common.yaml
+++ b/deployments/data100/config/common.yaml
@@ -103,6 +103,26 @@ jupyterhub:
         extraVolumeMounts:
           - name: home
             mountPath: /srv/homes
+#          - name: home
+#            mountPath: /home/jovyan/shared-readwrite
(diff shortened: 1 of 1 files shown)


felder (commented):
Lines 86-90 and 103-108 look suspect. 

Should we be doing 86-90 if we're trying to configure shared mounts based on course id? 

For that matter, the config starting at line 103 for admins seems suspect.  No subpath is defined (commented out) extra -homes commented out, etc.

balajialg:
@felder I have removed lines 86-90 and 103-108 as part of my latest commit.

felder:
@balajialg I thought we wanted to mount the root directory of _shared as readonly for hub admins so we could see everything stored there for all classes (not necessarily applicable to single course hubs, but should still be the same either way) when logging in.

PR #6839, opened 2025-02-24
pin the main hubploy repo

yijunge-ucb:


yijunge-ucb:
The workflow passed (https://github.com/berkeley-dsep-infra/datahub/actions/runs/13507856497/job/37741516651).

PR #6862, opened 2025-03-07
[DH-258] moving edx prod hub image over to edx GAR

felder:
https://jira-secure.berkeley.edu/browse/DH-258


felder:
This is a no op (edx is excluded from deployments), so I'm just gonna merge it.

PR #6907, opened 2025-03-21
[EdX] Add Kseniya to Admins

sean-morris:


felder:
@sean-morris Currently edx is excluded from the automated deployments via:

https://github.com/berkeley-dsep-infra/datahub/blob/1983d5bafa7703b0504fc91f330294142216166c/.github/workflows/deploy-hubs.yaml#L98

https://github.com/berkeley-dsep-infra/datahub/blob/1983d5bafa7703b0504fc91f330294142216166c/.github/workflows/deploy-hubs.yaml#L185

It's probably safe to remove this exclusion now that we've specified specific versions of the hub image for it to use...probably.  ;)

However this implies that any time a new hub image is built for edx, that it needs to go to GAR in the edx project and that staging.yaml and prod.yaml would need to be modified accordingly.  It also implies that this PR once merged will not be automatically deployed, so a follow up with a manual hubploy would be necessary (unless you've already done this) along with modification of the deployment workflow (probably as separate PR) if desired.

Other than that the PR itself looks fine.

felder (approved):

sean-morris:
@felder Thanks Jon. I think I get this:) -- a summary:
- I should submit a PR to get rid of the exclusions you specify
- If we build a new image(which this PR does not) the new imge tags need to be updated manually in staging and prod
- I need to run hubploy to deploy any change

felder:
@sean-morris 

Currently you need to run hubploy to deploy a change because edx is currently excluded from automated deployments.  If the exclusion is removed, automated deployments for edx will resume and you would not need to run hubploy manually.

If you build a new hub image specifically, those image tags need to be updated manually.  I believe the user-image tags will update.

PR #6909, opened 2025-03-25
[EdX] Removed Deployment Exclusions

sean-morris:
We specify the image tag for EdX hubs now so no reason to exclude the prod and staging hubs from deployment.


felder (changes requested):
@sean-morris why did you delete the entire line?  All you should've done was remove edx from both lines, not remove them entirely.

felder:
@sean-morris I'm going to close this PR and I'll submit another to make this change.

sean-morris:
@felder thanks Jon that was lame; I was not seeing cdss-discovery as a hub -- I don't what I thought it was

felder:
My PRs for this:  #6912 #6913

PR #6920, opened 2025-04-01
Github actions for automatic data retrieval

yijunge-ucb:


.github/scripts/sign_url_and_send_emails.py
```diff
@@ -0,0 +1,246 @@
...
+        print(f"Error: {e}")
+        return None
+
+
+def generate_html_body(urls):
+    # Create the HTML body
+    html_template = """
+    <html>
+    <head>
+        <title>Archived Files Retrieval</title>
```
balajialg, line 152: Can you include the 7 day limit for retrieval in the subject. Something like - Access Your Archived Files from DataHub (Valid for 7 Days)

balajialg (approved):
LGTM

.github/scripts/sign_url_and_send_emails.py
yijunge-ucb: The subject line has that information included "Access Your Archived Files from DataHub (Valid for 7 Days)". The title for the html template does not really do anything since this is displayed in an email, not an html web page.

balajialg: Sounds good

PR #7079, opened 2025-05-20
fix deprecation warning for ingress class

shaneknapp:
i tested this by manually deploying to `logodev` w/o issue.

i have the exact same configs in the cal-icor repo:  https://github.com/cal-icor/cal-icor-hubs/pull/103


ryanlovett (approved):
Makes sense, thanks!

PR #7100, opened 2025-05-24
Create a profile list from a user's courses.

ryanlovett:
This reads the user's courses from `auth_state`.

It runs an init container to precreate a subdirectory in their NFS home directory named after their Canvas course ID. It then mounts that subdirectory has their home on the hub.

I need to limit the list of courses to those that haven't ended yet, otherwise students will see all the course sites that they've ever been in.

I also need to enable the form to respond to people passing in something like `?profile=1234567&autostart=1`. `profile` would set the initial profile choice, and `autostart=1` would cause the server to launch. These would be necessary for nbgitpuller.

### DIFF
--- a/deployments/dev/config/common.yaml
+++ b/deployments/dev/config/common.yaml
@@ -26,7 +26,76 @@ jupyterhub:
       20-hubHome: |
         # Direct users to named server list
         c.JupyterHub.default_url = '/hub/home'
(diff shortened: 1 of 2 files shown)


ryanlovett:
I think I also need to add a "self-paced" profile so that students can do work that is not affiliated with a course.

PR #7274, opened 2025-07-17
Don't raise on 500.

ryanlovett:
The 500 codes we usually see are infrequent and harmless. Just log them.


shaneknapp (approved):
without getting too pedantic, checking for other errors is probly a good idea but maybe not for right now

PR #7275, opened 2025-07-17
update node placeholder helm chart

shaneknapp:
because i jumped the gun in https://github.com/berkeley-dsep-infra/datahub/pull/7274

### DIFF
--- a/node-placeholder/values.yaml
+++ b/node-placeholder/values.yaml
@@ -1,8 +1,7 @@
 replicaCount: 1
 
 image:
-  repository: 
-    us-central1-docker.pkg.dev/ucb-datahub-2018/core/node-placeholder-scaler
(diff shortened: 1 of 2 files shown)


node-placeholder/values.yaml
```diff
@@ -1,10 +1,11 @@
 replicaCount: 1
 
 image:
-  repository: us-central1-docker.pkg.dev/ucb-datahub-2018/core/node-placeholder-scaler
+  repository: 
+    us-central1-docker.pkg.dev/ucb-datahub-2018/core/node-placeholder-scaler
```
ryanlovett: This looks like a nested block rather than a `key: value` like in the previous version of the file. is chartpress being weird here?

ryanlovett (commented):
It seems like chartpress updated image.repository incorrectly? (putting it on two lines instead of one)

ryanlovett (approved):

PR #7308, opened 2025-07-23
[DH-548] Remove ugr entry in deploy-hubs GitHub action

balajialg:
Removes ugr associated entries in the yaml file to fix the issue with deploying to prod. Creating this PR based on Yijun's inputs!


yijunge-ucb (approved):

ryanlovett:
Should the ugr deployment just be deleted? Then it wouldn't have to be ignored.

yijunge-ucb:
@balajialg If we do not need to archive urg01, deleting deployments/urg01 from the github repo should fix the problem. Ryan is right, no need to ignore it.

balajialg:
@yijunge-ucb Sure, I am closing the PR. Can you please do the needful regarding deletion.

balajialg:
Okay, I removed the ugr01 directory from the repo and reverted the earlier entry ignoring the ugr01 hub during CI/CD builds

PR #7370, opened 2025-08-21
[DH-546] Make the shared directory visible in RStudio interface of the PH hub 

balajialg:
Hubployed the changes and tested with the GSI of the for the PH251BW course during the consultation

### DIFF
--- a/deployments/publichealth/config/common.yaml
+++ b/deployments/publichealth/config/common.yaml
@@ -148,6 +148,10 @@ jupyterhub:
             mountPath: /home/jovyan/_shared/bcourses-1545903-readwrite
             subPath: _shared/2025-fall/courses/1545903
             readOnly: false
+          - name: home
+            mountPath: /home/rstudio/_shared/bcourses-1545903-readwrite
+            subPath: _shared/2025-fall/courses/1545903
+            readOnly: false
       course::1545903::enrollment_type::ta:
       ## Course Name: bcourses-1545903 Bcourses ID: 1545903 End Date: 12/31/2025
       ## See issue https://github.com/berkeley-dsep-infra/datahub/issues/7357 for details.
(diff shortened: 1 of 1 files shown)


felder (changes requested):
Why are we mounting this to /home/jovyan and /home/rstudio, perhaps I'm mistaken but I don't think /home/jovyan exists on publichealth hub.

balajialg:
@felder, I can access `/home/jovyan` via terminal in PH hub. GitHub actions created the PR to add that particular path when the user submitted a shared directory request. I just pushed a commit to remove that stanza from the config. Let me know if it looks good to be merged.

felder:
@balajialg you can?

What am I missing?

```
rstudio@jupyter-felder:~$ echo $HOME
/home/rstudio
rstudio@jupyter-felder:~$ pwd
/home/rstudio
rstudio@jupyter-felder:~$ ls -l
total 4
drwxr-xr-x 6 rstudio rstudio 4096 Oct  9  2024 ph142-fa24
rstudio@jupyter-felder:~$ ls -al /home
total 12
drwxr-xr-x  1 root    root    4096 Oct 31  2024 .
drwxr-xr-x  1 root    root    4096 Aug 21 22:21 ..
drwxr-xr-x 11 rstudio rstudio 4096 Oct  9  2024 rstudio
rstudio@jupyter-felder:~$ 
```

balajialg:
@felder Thanks! I was able to access `/home/jovyan` in the instructor's terminal yesterday. I was able to access the `_shared` folder in `/home/jovyan` using their terminal (shared the snapshot below for your reference). However, When I tried the same in my terminal few minutes back, I couldn't find the directory (which aligns with your above point).

image

felder:
@balajialg ah ok, looking at that my guess is it was using /home/jovyan/… as a mount point that actually created the directory.  Yeah definitely best we don’t do that here.

balajialg:
@felder got it!

PR #7376, opened 2025-08-22
 Fall 25 setup: Implement bCourses roster-based authorization in per-course Data Hubs and increase node-placeholder value for hubs

balajialg:
Resolves [DH-546](https://jira-secure.berkeley.edu/browse/DH-546) and [DH-551](https://jira-secure.berkeley.edu/browse/DH-551)


felder (approved):
Looks good to me.

PR #7473, opened 2025-09-08
correct hub name parsing issues for hubs with dashes in names

yijunge-ucb:
correct hub name parsing issues for hubs with dashes in names


shaneknapp (approved):
tested locally and it works!

```
$ GITHUB_PR_LABEL_HUB_CDSS_DISCOVERY=1 .github/scripts/determine-hub-deployments.py --only-deploy cdss-discovery
cdss-discovery
```

PR #7585, opened 2025-09-29
DH-562

yijunge-ucb:
DH-562

### DIFF
--- a/docs/tasks/course-config.qmd
+++ b/docs/tasks/course-config.qmd
@@ -245,14 +245,17 @@ Instructors often request shared directories for collaboration with TAs and stud
             - "course::1111111"
     ```
 
-  This will generate Mounts 
+  This will generate the following mounts:
+
   for teachers and TAs:
+
   ```yaml
     - /home/jovyan/_shared/bcourses-1234567-readwrite 
     - /home/jovyan/_shared/bcourses-1234567
   ```
 
   for students
+
   ```yaml
     - /home/jovyan/_shared/bcourses-1234567
   ```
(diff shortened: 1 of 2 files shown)


yijunge-ucb:
@ryanlovett I think I've addressed the issues you mentioned in DH-562, but please feel free to share any feedback. Since the shared directory configurations are currently created automatically by GitHub Actions, this PR may need to wait until the corresponding updates are made there. That said, even if we merge this PR now, the existing configurations generated by GitHub Actions should still function correctly — it just might be a bit messy.

ryanlovett (commented):
I like that this includes automation to avoid the manual insertion of yaml! Regarding

```
custom:
   conda_shared:
     - "1234567, stat159, 8G"
```
> This will be translated into extraVolumeMounts, mem_limit, mem_guarantee like:

You might consider changing it to:

```
custom:
   conda_shared:
     - bcourse_id: 1234567
       staff_overrides:
         mem_limit: 8G
         mem_guarantee: 8G
 ```

This way you don't have to parse it again, and it is more readable.

I expanded `8G` into `staff_overrides` to be synonymous with our existing overrides as well as kubespawner's, even though there probably won't be any other fields that get overridden due to `conda_shared`. It also might be necessary to make limit/guarantee different in some cases.

I removed the hub name -- can that be obtained from the deployment the configuration is specified in?

Also, do you need to translate this into our custom config first? i.e. assembling "::" delimited strings and then parsing them later. Can you instead convert `conda_shared` directly into kubespawner overrides?

yijunge-ucb:
@ryanlovett I made some changes according to your comments.

ryanlovett (commented):
Thanks @yijunge-ucb ! Can you please also add some information about this to the docs and maybe add some function docstrings?

yijunge-ucb:
Thanks @ryanlovett! Docstrings and docs for shared conda and shared directories configs are added.

ryanlovett (approved):
Looks good!

PR #7592, opened 2025-10-01
Merge 7396 and 7590 to prod

balajialg:
@shaneknapp @minrk Can your changes get merged to prod?


shaneknapp:
i believe so, yes...  @minrk are these the correct image hashes for stat159?

minrk:
As I understand it, #7590 prevents stat159 updating to prod, is that right? If so, this should be fine.

shaneknapp:
> As I understand it, #7590 prevents stat159 updating to prod, is that right? If so, this should be fine.

yep!

@balajialg i'll merge this for ya

balajialg:
Thank you both!

PR #7688, opened 2025-12-01
Set windowing mode to `defer` in Data8 hub

balajialg:
Ref: https://github.com/berkeley-dsep-infra/datahub/pull/7434/files

@felder fixed a similar scrolling issue reported by instructors using R and Data100 hubs at the start of FA 25.


balajialg:
@yijunge-ucb approved the changes. Merging the PR to staging

PR #7874, opened 2026-01-20
moving to pinned and working version of `get-labels-on-push`

shaneknapp:
tested this out last week and everything works as advertised!

setting the version to `v1` as per the developer's comment [here](https://github.com/irby/get-labels-on-push/issues/7#issuecomment-3769737915)


yijunge-ucb:
Thanks so much @shaneknapp! actions/github-script might be able to extract labels too, but I need to figure out the details. If it works, we might want to use it instead since it is the official Github action.

PR #8060, opened 2026-02-23
[DH-620] Redact email ids in GitHub Issue description and comments

balajialg:
Just created the PR for testing purposes. Won't merge it to the staging branch.


.github/workflows/data-retrieval.yaml
```diff
@@ -236,11 +236,95 @@ jobs:
...
+                await github.rest.issues.updateComment({
+                  owner, repo,
+                  comment_id: comment.id,
+                  body: redactedCommentBody
+                });
+                redactedCount++;
+              }
+            }
+
+            if (redactedCount > 0) {
```
yijunge-ucb, line 316: If emails were adapted, a label "delete-email" should be added to the issue so that a following workflow can be triggered.

yijunge-ucb: Can you link me to the test workflows too? With original code (exact code like this file except for missing the email sending steps), issue created, and workflow triggered. The workflow you sent through slack https://github.com/balajialg/dataretrieval_improvements/actions/runs/22376266608 is using a different workflow file than this one.

yijunge-ucb (changes requested):

.github/workflows/data-retrieval.yaml
balajialg: Are we doing the "delete-email" label in this iteration of the workflow? Didn't we align that is out of scope for iteration 1? If you want that as part of version 1, I can update the PR accordingly. 

1. GitHub Issue Created: https://github.com/balajialg/dataretrieval_improvements/issues/31
2. GitHub Action Triggered: https://github.com/balajialg/dataretrieval_improvements/actions/runs/22598954378/job/65476030200
3. YAML file triggering the workflow - https://github.com/balajialg/dataretrieval_improvements/blob/main/.github/workflows/data-retrieval.yaml

I uncommented the email-sending workflow from that file and made tweaks to confirmation message.

yijunge-ucb: Adding the label step goes into the same file, and same PR. If you prefer merging one PR first, creating another PR, adding one step in the same file, then merging that PR, I am ok with that too. What is the timeline for this issue? Do you prefer me doing this if the timeline is not too strict?

balajialg: @yijunge-ucb There is no strict timeline for this task, and it is up to the two of us to prioritize this task and ensure that it doesn't fall through the cracks. I took it up since I thought that with the help of AI agents, I might find a good enough solution for this issue (& ensure that there is one less task on you). However, you might come up with a far more elegant solution than what is implemented here. If you want to take it up during the summer when we don’t have much activity, that sounds good to me. Otherwise, I can update the PR with the label workflow. Let me know!

yijunge-ucb: I do not have a more elegant solution, and I would probably do more or less the same. This is definitely a good start. I can work on the rest of the issue since I was the one that set up the secrets in the datahub repo and it is probably easier for me to test things.

balajialg: Sounds like a plan. For my completion sake, I added the workflow to add the "delete-email" label after GitHub Action posts the comment. 

1. Test GitHub Issue - https://github.com/balajialg/dataretrieval_improvements/issues/32
2. Commit - https://github.com/balajialg/dataretrieval_improvements/commit/f5b8e799d77b53157c12da08756cff89e248897a
3. GitHub Action - https://github.com/balajialg/dataretrieval_improvements/actions/runs/22603194086

PR #8167, opened 2026-04-08
Update course configuration in common.yaml

yijunge-ucb:
DH-672, to address the request https://github.com/berkeley-dsep-infra/datahub/issues/7778


balajialg:
Merging the changes since the instructor acknowledged it

PR #8183, opened 2026-04-16
[Edx] Removed Edx From Deployment

sean-morris:
- All artifacts related to EdX via a search of the datahub directory are removed
- All references to Edx are removed from remaining files
- Just for the record I did a search of ucb-datahub-2018 GCP project and found nothing related to EdX hanging around:
  - GKE node pools
  - Compute disks
  - Cloud Storage buckets
  - Service accounts


yijunge-ucb (approved):
Looks good to me!

PR #8304, opened 2026-06-16
[DH-688] Update bCourses authorization list in Data8, Stat20 and Data100 hubs for Summer 2026

balajialg:


pancakereport (approved):
DATA course IDs are correct

